PR #13, opened 2026-02-10
Don't overwrite default model with `blockrun/auto`

achiurizo:
When using the `openclaw` to handle installation of `clawrouter`, you can run into an issue where `blockrun/auto` is set as the model and this creates connection issues with the model as the process is still being setup. This requires manual intervention to get back on track.

By not overriding the model, the `openclaw` agent can continue the process until completion and instead we prompt the user or agent to switch via `/model blockrun/auto`

### DIFF
--- a/scripts/reinstall.sh
+++ b/scripts/reinstall.sh
@@ -165,7 +165,10 @@ if (fs.existsSync(configPath)) {
 "
 
 echo ""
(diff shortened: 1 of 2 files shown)


achiurizo:
I tried with latest and did a reinstall and the flow works better now. Thank you!

PR #15, opened 2026-02-10
feat: Local Claude CLI integration to skip x402 for Anthropic models

### LINKED ISSUE #14: Configure to allow local Claude CLI instance with Pro/Max subscription
the anthropic API is more expensive at a certain point compared to the Pro/Max subscriptions. I have Claude CLI installed and authenticated on my machine but I'd prefer to only use it for coding tasks and use ClawRouter for routing everything else. This doesn't seem to be an option at the moment but it would be an amazing feature to have.

gigi-trifle:
Implement the feature from issue #14. Adds config option to route Anthropic models to local Claude CLI instance instead of x402 payments. Includes selector logic, proxy handler, and config injection. POC tested with basic requests. Closes #14.


1bcMax:
Hi @gigi-trifle,

Thank you for taking the time to implement this feature! The code quality is solid and the approach is clever. However, after careful consideration, we've decided not to merge this PR. Let me explain why:

## Technical & Compliance Concerns

**1. Terms of Service Risk**  
Claude Code's subscription ToS may not permit usage through proxy services. By integrating this directly into ClawRouter, we'd be facilitating potential ToS violations, which puts both BlockRun and users at risk.

**2. Security Considerations**  
The implementation requires:
- Spawning external processes (Claude CLI)
- File system access for CLI execution
- Managing user credentials/sessions

This increases ClawRouter's attack surface significantly. It's better to keep authentication at the OpenClaw layer where users have full control.

**3. Maintenance & Scope Creep**  
Supporting Claude CLI is just the beginning. Users would then request:
- ChatGPT Plus support (different auth method)
- Gemini Advanced support (OAuth flow)
- Compatibility with all CLI versions
- Platform-specific issues (Mac/Linux/Windows)

This shifts ClawRouter's focus from "cost-optimized routing" to "subscription management platform."

## The Alternative: OpenClaw-Native Approach

OpenClaw **already has a model failover system** that solves this use case more elegantly:

```bash
openclaw models set anthropic/claude-sonnet-4-5  # Primary (subscription)
openclaw models fallbacks add blockrun/auto       # Failover (x402)
```

**Advantages over this PR:**
- ✅ Works with **any provider** (not just Claude)
- ✅ No ClawRouter code changes needed
- ✅ Respects provider ToS (user configures auth directly)
- ✅ Zero maintenance burden
- ✅ More flexible (supports complex fallback chains)

## Documentation Instead of Code

We've created comprehensive documentation explaining how to set this up:

📖 **[Subscription Failover Guide](https://github.com/BlockRunAI/ClawRouter/blob/feature/subscription-failover/docs/subscription-failover.md)**

This guide covers everything users need to achieve the same goal without the compliance/security risks.

## Feedback Welcome

We really appreciate your contribution and the discussion it sparked! If you have thoughts on the OpenClaw-native approach or see gaps we haven't addressed, please share them here or in the original issue (#14).

Thank you again for your interest in improving ClawRouter!

**Resources:**
- 📖 [Setup Guide](https://github.com/BlockRunAI/ClawRouter/blob/feature/subscription-failover/docs/subscription-failover.md)  
- 💬 [Telegram Community](https://t.me/blockrunAI)

PR #34, opened 2026-02-16
fix: handle wrapped x402 premium payment failures with safer fallback

tsubasakong:
## Summary
- treat wrapped x402 merchant-proxy payment failures as provider errors that can fallback
- improve session/profile fallback routing to avoid sticky premium routing after paid-model failure
- add regression tests for wrapped x402 error fallback and profile re-routing

## Test Plan
- [x] `bun run test/premium-compat.ts`
- [x] `npm run typecheck`

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -287,6 +287,10 @@ const PROVIDER_ERROR_PATTERNS = [
   /billing/i,
   /insufficient.*balance/i,
   /credits/i,
+  /payment required/i,
+  /x402[_-]?payment[_-]?failed/i,
+  /merchant_status\"?\s*:\s*402/i,
+  /endpoint requires x402 payment/i,
   /quota.*exceeded/i,
   /rate.*limit/i,
   /model.*unavailable/i,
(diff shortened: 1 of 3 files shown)


VickyXAI:
Thanks — closing the diff, but the underlying question is worth keeping.

Stale: main's `setSession()` 4th param is now `userExplicit: boolean`, not `routingProfile`, so this would pass a profile string into the explicit-pin slot. `SessionEntry` has since grown three-strike escalation and `sessionCostMicros`.

Also, one part I'd push back on regardless of age: adding `/payment required/i` and `/merchant_status\"?\\s*:\\s*402/i` to `PROVIDER_ERROR_PATTERNS` reclassifies "your wallet is out of money" as "the provider is flaky", and `/payment required/i` is broad enough to match model *output* text. Main deliberately separates these — it has dedicated `invalid_payload` and `insufficient_funds` branches with specific guidance — so this would mask real payment failures behind a silent reroute.

The bug you found may well still be live though: if a `premium → eco` profile switch still leaves a session pinned to the paid model, please open an issue with a repro and we'll fix it against the current session code.

PR #35, opened 2026-02-16
feat: add claw credit payment mode integration with setup tooling

tsubasakong:
## Summary
- add `paymentMode: clawcredit` flow for BlockRun inference without local wallet signing
- switch claw.credit pay path to official `@t54-labs/clawcredit-sdk`
- add OpenClaw setup helper script: `scripts/setup-clawcredit.sh`
- include bash-safety fixes for the setup script

## Test Plan
- [x] `npx -y tsx test/clawcredit-mode.ts`
- [x] `npm run typecheck`
- [x] `bun run build`

### DIFF
--- a/README.md
+++ b/README.md
@@ -292,6 +292,18 @@ For basic usage, no configuration needed. For advanced options:
 | `CLAWROUTER_DISABLED` | `false` | Disable smart routing |
 | `BLOCKRUN_PROXY_PORT` | `8402`  | Proxy port            |
 | `BLOCKRUN_WALLET_KEY` | auto    | Wallet private key    |
+| `BLOCKRUN_PAYMENT_MODE` | `wallet` | Payment backend: `wallet` or `clawcredit` |
+| `CLAWCREDIT_API_TOKEN` | - | Required when `BLOCKRUN_PAYMENT_MODE=clawcredit` |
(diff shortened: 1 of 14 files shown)


VickyXAI:
Thanks for the work here, and sorry it sat this long — but I'm closing this on security grounds, independent of the staleness.

**`headersToObject()` forwards credentials to a third party.** It strips only `host`, `content-length` and `connection`, so `Authorization` and every upstream API-key header are copied into `request_body.http.headers` and sent to `api.claw.credit`. The `openclaw.security.json` entry discloses the bearer token as `sent-to-claw-credit-api` but not the forwarded upstream credentials. A header **allowlist** is the only safe shape here, not a denylist.

Two more, worth fixing in any future version:
1. **Fabricated success** — the SDK path returns `new Response(JSON.stringify(merchant_response ?? {}), { status: 200 })`, an unconditional 200 even when `merchant_response` is undefined (becomes `{}`). Upstream failures get laundered into apparent success.
2. **Supply chain** — `@t54-labs/clawcredit-sdk@^0.2.40` is a floating `^0.x` range, and tsup's `noExternal: [/.*/]` bundles it into our published artifact. Also `microsToUsd` defaults to $0.01 on unparseable input rather than failing.

Not a judgement on claw.credit — a credit rail is a reasonable thing to want. But it needs a header allowlist, real status propagation, and a pinned/audited dep, which is a fresh design rather than a rebase.

PR #58, opened 2026-02-26
feat: add CDP (Coinbase Developer Platform) MPC wallet support

skingem1:
## CDP Wallet Support for ClawRouter

Adds Coinbase Developer Platform (CDP) MPC wallets as an alternative to raw private keys for x402 payments.

### Why?
- **MPC-based** — no single private key to leak or lose
- **Enterprise-grade** — Coinbase-managed backup and recovery
- **Agent-native** — programmatic wallet creation via API
- **Fully backward compatible** — raw private key still works exactly as before

#### New files
- `src/wallet-signer.ts` — `WalletSigner` interface (abstract signing layer)
- `src/wallet-cdp.ts` — CDP MPC wallet implementation using `@coinbase/coinbase-sdk`
- `src/wallet-pk.ts` — Raw private key wrapper (satisfies `WalletSigner`)
- `docs/cdp-wallet.md` — Full setup documentation

#### Updated files
- `src/x402.ts` — `createPaymentFetch()` now accepts `WalletSigner | `0x${string}``
- `src/proxy.ts` — `ProxyOptions` accepts optional `walletSigner`
- `src/index.ts` — Auto-detects CDP env vars at startup, falls back to raw key
- `src/auth.ts` — New `cdpWalletAuth` (wizard) + `cdpEnvAuth` (env vars) methods

### Usage

```bash
# Set CDP credentials
export BLOCKRUN_CDP_API_KEY_NAME="organizations/xxx/apiKeys/yyy"
export BLOCKRUN_CDP_PRIVATE_KEY="-----BEGIN EC PRIVATE KEY-----\n..."

# ClawRouter auto-detects and uses CDP wallet on startup
openclaw gateway restart
```

### Build
- ESM ✅
- DTS ✅
- Zero breaking changes

### DIFF
--- a/docs/cdp-wallet.md
+++ b/docs/cdp-wallet.md
@@ -1,0 +1,80 @@
+# CDP Wallet Support
+
+ClawRouter supports **Coinbase Developer Platform (CDP) MPC wallets** as an alternative to raw private keys for x402 payments.
+
+## Why CDP?
+
+| Feature | Raw Private Key | CDP MPC Wallet |
+|---------|----------------|----------------|
+| Key storage | Single file on disk | Distributed MPC — no single key |
+| Recovery | Manual backup required | Coinbase-managed recovery |
+| Setup | Auto-generated | Requires CDP API key |
+| Best for | Individual use, quick start | Teams, production agents |
+
+## Setup
+
+### 1. Create a CDP API Key
+
+1. Go to [portal.cdp.coinbase.com](https://portal.cdp.coinbase.com)
+2. Create a new API key
+3. Download the API key name and private key
+
+### 2. Configure ClawRouter
+
+**Option A — Environment variables (recommended for production):**
+
+```bash
+export BLOCKRUN_CDP_API_KEY_NAME="organizations/xxx/apiKeys/yyy"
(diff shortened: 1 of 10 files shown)


1bcMax:
Thank you for your suggestion. let me think about it.

VickyXAI:
Thanks for this, and apologies it sat. Closing as not-rebasable rather than not-wanted: it rewrites `src/x402.ts` and changes `createPaymentFetch()`'s signature — both of which no longer exist. Main now builds its pay-fetch from `x402Client` (@x402/fetch) + `createPayFetchWithPreAuth`, and `toClientEvmSigner` from @x402/evm *is* the signer abstraction now. So the seam you wanted exists; it just landed differently.

Two things worth flagging if you or anyone revisits CDP support:
1. `createAndPersistWallet` calls `wallet.export()` and writes the seed in plaintext to `~/.openclaw/blockrun/cdp/wallet.json` — which defeats the stated premise of "no single private key to leak".
2. The EIP-712 path hashes with viem then calls `defaultAddress.signHash(hash)`; there's no test proving a CDP-signed authorization actually settles with the facilitator (v normalization / format).

A fresh PR against @x402/evm's signer interface would be welcome.

PR #97, opened 2026-03-16
feat: add wzrd-momentum skill — AI adoption velocity for routing

twzrd-sol:
Adds WZRD momentum skill for model routing.


1bcMax:
Thanks for the contribution\! However, we're not accepting external service integrations as skills at this time. Our routing decisions are based on internal performance data and user retention metrics. Closing for now.

PR #104, opened 2026-03-17
fix: Timeout (AbortError) triggers fallback chain instead of skipping

### LINKED ISSUE #95: Timeout (AbortError) does not trigger fallback model chain
## Summary

When an LLM request hits the hard timeout (`AbortError`), the extension throws an error **outside** the fallback `for` loop, causing the full fallback model chain to be silently skipped. No fallback model is ever tried on timeout — only on provider errors (4xx/5xx).

## Expected Behavior

On `AbortError` (request timeout), the loop should `continue` to the next model in the fallback chain — the same behavior that occurs when a provider returns a 4xx/5xx error.

## Actual Behavior

The `throw` in the `AbortError` handler (around `proxy.ts` line 1102) exits the try/catch and the entire `for` loop. The full fallback chain configured in `openclaw.json` is never invoked on timeouts.

## Steps to Reproduce

1. Configure a multi-model fallback chain in `openclaw.json`
2. Send a high-complexity request (long context + image attachments) that exceeds the `DEFAULT_REQUEST_TIMEOUT_MS` threshold
3. Observe: the primary model times out, no fallback model is attempted, the request fails with a hard `FailoverError`

## Root Cause

In `proxy.ts`, the `AbortError` catch block throws unconditionally:
```js
if (err instanceof Error && err.name === "AbortError") throw new Error(`Request timed out...`)
```

This `throw` exits the fallback `for` loop entirely instead of `continue`-ing to the next model.

## Suggested Fix

Move the timeout handler inside the fallback loop so it can `continue` to the next model:

```js
// Inside the fallback for loop
if (err instanceof Error && err.name === "AbortError") {
  console.warn(`Model ${modelId} timed out, trying next fallback...`);
  continue; // instead of throw
}
```

## Environment

- OpenClaw gateway with ClawRouter extension
- Fallback chain: `claude-max/claude-sonnet-4` → `claude-max/claude-opus-4` → `openai-codex/gpt-5.3-codex` → (additional fallbacks)
- Hard timeout: `DEFAULT_REQUEST_TIMEOUT_MS` (~180–300s)
- Observed: 7+ timeout failures in a single session, all skipping the fallback chain

## Impact

Users relying on fallback chains for reliability get hard failures on timeout instead of automatic model fallover. This defeats the purpose of the fallback configuration.

Jah-yee:
## Summary

When an LLM request hits the hard timeout (`AbortError`), the extension was throwing an error **outside** the fallback `for` loop, causing the full fallback model chain to be silently skipped. No fallback model was ever tried on timeout — only on provider errors (4xx/5xx).

## Fix

- Wrap `tryModelRequest` in try/catch inside the fallback loop
- On AbortError (timeout), log warning and continue to next model
- This fixes the issue where timeouts would exit the fallback loop entirely
- Previously, only provider errors (4xx/5xx) triggered fallback; now timeouts do too

## Changes

- `src/proxy.ts`: Added try/catch around `tryModelRequest` call inside the fallback for loop
- On AbortError, sets `lastError` with timeout message and continues to next model if not last attempt

Fixes #95

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timeout handling to gracefully fall back to alternative models instead of immediately failing when a request times out.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -3423,17 +3423,34 @@ async function proxyRequest(
 
       console.log(`[ClawRouter] Trying model ${i + 1}/${modelsToTry.length}: ${tryModel}`);
 
-      const result = await tryModelRequest(
-        upstreamUrl,
-        req.method ?? "POST",
-        headers,
-        body,
-        tryModel,
-        maxTokens,
-        payFetch,
-        balanceMonitor,
-        controller.signal,
-      );
+      let result: ModelRequestResult;
+      try {
+        result = await tryModelRequest(
+          upstreamUrl,
+          req.method ?? "POST",
+          headers,
+          body,
+          tryModel,
+          maxTokens,
+          payFetch,
+          balanceMonitor,
+          controller.signal,
+        );
+      } catch (err) {
+        // Timeout (AbortError) should trigger fallback to next model, not exit the loop
+        if (err instanceof Error && err.name === "AbortError") {
+          console.log(`[ClawRouter] Timeout on ${tryModel}, trying fallback...`);
+          lastError = {
+            body: `Request timed out after ${timeoutMs}ms`,
+            status: 408,
+          };
+          if (!isLastAttempt) {
+            continue;
+          }
+        }
+        // Re-throw other errors to be handled by outer catch
+        throw err;
+      }
 
       if (result.success && result.response) {
         upstream = result.response;


1bcMax:
Closing — the timeout fallback issue has already been fixed on main with a better approach (per-model AbortController + global/per-model two-tier timeout). This PR is based on older code and has merge conflicts.

PR #109, opened 2026-03-21
feat: support fxUSD and multi-decimal EIP-3009 stablecoins

huwangtao123:
## What
Add proper multi-decimal EIP-3009 stablecoin support, fixing bugs that would cause incorrect behavior with non-6-decimal assets like fxUSD (18 decimals).

## Why
ClawRouter's payment layer was hardcoded to assume 6 decimals (USDC) in several places. Adding fxUSD (18 decimals, [verified on Basescan](https://basescan.org/token/0x55380fe7a1910dff29a47b622057ab4139da42c5)) exposed these gaps.

## Changes
- **`/health` endpoint** now returns `paymentAssetDecimals`
- **Proxy reuse path** preserves `decimals` from existing proxy (fixes 10^12× balance inflation for 18-decimal assets)
- **`transformPaymentError`** uses asset-native decimals for USD conversion in `insufficient_funds` errors
- **`payment-asset.ts`** (new) — `BasePaymentAsset` type, normalizer, fetcher
- **`balance.ts`** — updated USDC-specific comments to generic stablecoin
- **3 new tests** — fxUSD normalization, mixed-decimal assets, real contract address

## On-Chain Verification
fxUSD contract (`0x55380fe7a1910dff29a47b622057ab4139da42c5`) on Base implements full EIP-3009:
- `transferWithAuthorization` ✅
- `receiveWithAuthorization` ✅
- `cancelAuthorization` ✅
- All typehashes + `authorizationState` + `DOMAIN_SEPARATOR` ✅

## Testing
- 363 tests pass, 0 regressions
- 3 new tests added for fxUSD-specific scenarios

## Summary by CodeRabbit

* **New Features**
  * Multi-asset Base support with per-request asset selection, runtime asset switching, normalized USD-micros balances, and payment-asset metadata surfaced in health/proxy outputs; default remains USDC on Base.
* **Bug Fixes / UX**
  * Balance checks, shortfall messages and funding prompts now show the active asset symbol and use consistent USD formatting.
* **Tests**
  * New tests covering payment-asset normalization, fetching, ordering and fallback behavior.


VickyXAI:
Thanks for this — and to be clear up front, the math is *correct*. `toUsdMicros` (`rawAmount / 10n ** BigInt(decimals - 6)`) is BigInt throughout and truncates conservatively, and I checked that the signed amount still comes from the server's `selectedRequirements.amount`, so there was never a 10^n mispay risk here. The decimals work only affects balance pre-checks, error text and /health.

Closing for a different reason: it's client-side support for a server capability that doesn't exist. It discovers assets via `GET /v1/payment-metadata?chain=base`, which returns 404 on the live gateway today. A client can only pay with what the server advertises in its 402, and the gateway advertises USDC only — so `fetchBasePaymentAssets` always falls through to `DEFAULT_BASE_PAYMENT_ASSET`. The 1656 lines unconditionally no-op.

This is server-led. If we ship multi-asset 402s, this is worth reopening against that work — but the diff is 4 months old, conflicting, and includes 730 lines of regenerated dist/, so it wouldn't survive the rebase anyway. Sorry it sat this long without that being said.

PR #140, opened 2026-04-02
fix(plugin): wire up pluginConfig.walletKey declared in configSchema

Reblexis:
## Bug

`openclaw.plugin.json` declares `walletKey` in its `configSchema`:

\`\`\`json
"walletKey": {
  "type": "string",
  "description": "EVM wallet private key (0x...). Optional — auto-generated if not set."
}
\`\`\`

The programmatic `startProxy({ walletKey })` API also accepts it. However, `startProxyInBackground()` never reads `api.pluginConfig?.walletKey` — it always calls `resolveOrGenerateWalletKey()` (saved file → env var → auto-generate), silently ignoring any value set in `openclaw.json`.

## Fix

Check `api.pluginConfig?.walletKey` before the file/env resolution chain. If present and valid, use it directly with source `"config"`. The existing resolution order is unchanged when `walletKey` is absent or invalid.

\`\`\`diff
-const wallet = await resolveOrGenerateWalletKey();
+const configKey = api.pluginConfig?.walletKey as string | undefined;
+let wallet: WalletResolution;
+if (typeof configKey === "string" && configKey.startsWith("0x") && configKey.length === 66) {
+  const account = privateKeyToAccount(configKey as \`0x\${string}\`);
+  wallet = { key: configKey, address: account.address, source: "config" };
+} else {
+  wallet = await resolveOrGenerateWalletKey();
+}
\`\`\`

## Testing

Added `walletKey` to the `plugins.entries.clawrouter.config` section of `openclaw.json`, restarted the gateway, and confirmed the log line:

\`\`\`
[gateway] Using wallet from plugin config: 0x97d9...
\`\`\`

Previously the gateway always logged `Using saved wallet: 0x<treasury-wallet>` regardless of any configured key.

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Plugin config can now supply a wallet private key; when valid, it is used ahead of saved wallets and environment variables.
  * Logging now indicates when a wallet from plugin config is in use, alongside existing source notices.

* **Bug Fixes**
  * If a configured wallet key is invalid, a warning is logged and the system falls back to the previous resolution order.

### DIFF
--- a/src/auth.ts
+++ b/src/auth.ts
@@ -215,7 +215,7 @@ async function generateAndSaveWallet(): Promise<{
 export type WalletResolution = {
   key: string;
   address: string;
-  source: "saved" | "env" | "generated";
+  source: "saved" | "env" | "config" | "generated";
   mnemonic?: string;
   solanaPrivateKeyBytes?: Uint8Array;
 };
--- a/src/index.ts
+++ b/src/index.ts
@@ -33,6 +33,7 @@ import {
   WALLET_FILE,
   MNEMONIC_FILE,
 } from "./auth.js";
+import type { WalletResolution } from "./auth.js";
 import type { RoutingConfig } from "./router/index.js";
 import { BalanceMonitor } from "./balance.js";
 import {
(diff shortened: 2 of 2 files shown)


Reblexis:
Addressed in the second commit: switched to `/^0x[0-9a-fA-F]{64}$/.test(configKey)` (same regex used in `walletKeyAuth`) and added a warning log when the key is present but invalid, so misconfiguration is visible rather than silent.

Issue #147, opened 2026-04-09
Custom routing config in openclaw.json is silently ignored


jeegankones:
I defined custom routing tiers in `openclaw.json` under `plugins.entries.clawrouter.config.routing`, but the proxy always starts with the default routing config. No errors or warnings, it just silently uses defaults.

After digging into the source, I found the cause: `register(api)` gets called twice during OpenClaw's gateway startup. The first call happens in a pre-gateway phase before `openclaw.json` has been read, so `api.pluginConfig` is empty. The proxy starts with `DEFAULT_ROUTING_CONFIG` at this point. Then `proc.__clawrouterProxyStarted` gets set to `true`, which causes the second `register()` call (the one that actually has the user's config in `api.pluginConfig`) to bail out immediately.

You can see this in the gateway log: the timestamp-less `[plugins]` lines (pre-gateway phase) appear before `[gateway] loading configuration…`, and all subsequent timestamped calls log "Proxy already started by earlier register() call - skipping".

This means `plugins.entries.clawrouter.config.routing` effectively does nothing for anyone.

**Repro**: Define any custom tiers in `openclaw.json`, restart the gateway, check `blockrun/logs/usage-*.jsonl`. Routed models will match `DEFAULT_ROUTING_CONFIG`, not the user config.

**Version**: `@blockrun/clawrouter@0.12.139`, OpenClaw `2026.4.9`

PR #149, opened 2026-04-10
fix: apply routing config on proxy reuse path (defense-in-depth)

kagura-agent:
## Summary

Defense-in-depth fix for the `startProxy()` reuse path: when a second `register()` call hits an existing proxy via `checkExistingProxy()`, `routingConfig` was silently dropped because `mergeRoutingConfig` was never called on that path.

**Note:** This is a *different* bug than #147 (which was fixed in v0.12.142 via the single-process plugin path). This PR addresses the multi-process reuse scenario where `startProxy()` returns an existing proxy URL.

### What changed

1. New `/__update-routing` internal endpoint on the proxy server — accepts POST with routing config and calls `mergeRoutingConfig` on the running instance
2. `startProxy()` reuse path sends routing config to the running proxy when `routingConfig` is present
3. New `/__routing-config` debug endpoint for inspection
4. Returns `{ updated: false }` when no `routingConfig` provided (callers can distinguish no-op vs actual update)

### Limitation

Config applied this way is *additive* — it cannot remove overrides set by the first caller. This is inherent to the reuse model.

### Tests

- 2 new tests for the reuse-path scenario
- All 410 existing tests pass
- TypeCheck clean

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -516,9 +516,6 @@ function injectModelsConfig(
   }
 }
 
-function readStringSafe(value: unknown): string | undefined {
-  return typeof value === "string" && value.trim() ? value.trim() : undefined;
-}
 
 /**
  * Inject dummy auth profile for BlockRun into agent auth stores.
--- a/src/proxy.routing-config-reuse.test.ts
+++ b/src/proxy.routing-config-reuse.test.ts
@@ -49,7 +49,9 @@ describe("startProxy routing config reuse", () => {
       expect(updatedConfig.tiers.SIMPLE.fallback).toEqual(["test-fallback-1"]);
 
       // Other tiers should still have defaults (merged, not replaced)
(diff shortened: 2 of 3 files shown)


kagura-agent:
@coderabbitai Valid point — fixed in e6c4181. The endpoint now returns `{ updated: false }` when `body.routingConfig` is falsy, so callers can distinguish between an actual config update and a no-op.

kagura-agent:
Thanks for the thorough breakdown @1bcMax! You're right — the actual #147 flow never hits the reuse path, so the framing was off.

I'll go with **(b)**: revise the description to frame it as a defense-in-depth fix for the `startProxy()` reuse path (not #147), and note the limitation re: immutable config (`walletKey`, `maxCostPerRun`, etc.). The `__routing-config` GET endpoint is a nice debugging bonus.

Will push the update shortly.

kagura-agent:
Fixed lint error — removed unused `readStringSafe` function.

PR #150, opened 2026-04-10
fix: merge agenticTiers/ecoTiers/premiumTiers in routing config overrides

### LINKED ISSUE #148: Custom tiers have no effect on agent requests because agenticTiers isn't user-configurable
Even after getting custom routing config to load (see #147), I noticed all agent requests still route to the default `agenticTiers` models (kimi-k2.5, grok) instead of my custom tier definitions.

The problem is twofold:

1. `mergeRoutingConfig` deep-merges `tiers`, `classifier`, `scoring`, and `overrides` from user config, but doesn't touch `agenticTiers`, `ecoTiers`, or `premiumTiers`. So there's no way for a user to customize which models handle agentic requests.

2. Every agent request includes tools, which means `hasToolsInRequest` is always `true`. This makes `useAgenticTiers = true` whenever `config.agenticTiers != null`, which it always is since it comes from the defaults. Setting `overrides.agenticMode: false` doesn't help because it only controls `isExplicitAgentic`, which is ORed with `hasToolsInRequest`. The name `agenticMode` suggests it's a toggle for the whole agentic tier system, but it isn't.

The practical effect: defining custom `tiers` in the config does nothing for the actual agent workload, since all agent requests go through `agenticTiers` which can't be overridden. The only workaround I found was setting `"agenticTiers": null` in the routing config to force everything through regular `tiers`.

**Suggestion**: `mergeRoutingConfig` should merge all tier sets the same way it merges `tiers`, and there should be a clear way to disable agentic tier selection entirely (or `agenticMode: false` should actually do that).

**Version**: `@blockrun/clawrouter@0.12.139`

kagura-agent:
## Problem

`mergeRoutingConfig` only deep-merges `tiers`, `classifier`, `scoring`, and `overrides` from user config. The three specialized tier sets (`agenticTiers`, `ecoTiers`, `premiumTiers`) are silently ignored, making it impossible to customize which models handle agentic/eco/premium requests.

Since every agent request includes tools (`hasToolsInRequest = true`), `agenticTiers` is always used for agent workloads. Custom `tiers` definitions have no effect on actual agent traffic.

## Fix

Added `mergeTierRecord()` helper with three-case handling:
- **Not provided** (`undefined`) → keeps default
- **Explicitly `null`** → disables that tier set (sets to `undefined`)  
- **Object** → deep-merges with default

Applied to all three optional tier sets in `mergeRoutingConfig()`.

## Tests

13 new tests covering:
- `mergeTierRecord` base cases (null, undefined, merge, no-default)
- Custom `agenticTiers` override merges correctly
- `agenticTiers: null` disables agentic tier selection
- `ecoTiers`/`premiumTiers` custom overrides
- Combined overrides with other routing config fields

All 409 tests pass. TypeCheck clean.

Fixes #148

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive tests for routing-tier merging: default preservation, explicit disabling, shallow tier-level merges, and interactions with other routing fields.

* **Chores**
  * Updated routing configuration merging to support explicit disabling of tier maps, preserve defaults when unspecified, and correctly merge partial tier maps so only specified entries change.

### DIFF
--- a/src/proxy.merge-routing.test.ts
+++ b/src/proxy.merge-routing.test.ts
@@ -66,9 +66,7 @@ describe("mergeRoutingConfig", () => {
     });
     expect(result.agenticTiers!.SIMPLE).toEqual(customSimple);
     // Other tiers preserved from default
-    expect(result.agenticTiers!.COMPLEX).toEqual(
-      DEFAULT_ROUTING_CONFIG.agenticTiers!.COMPLEX,
-    );
+    expect(result.agenticTiers!.COMPLEX).toEqual(DEFAULT_ROUTING_CONFIG.agenticTiers!.COMPLEX);
   });
 
   it("disables ecoTiers when set to null", () => {
@@ -105,9 +103,7 @@ describe("mergeRoutingConfig", () => {
       },
     });
     expect(result.premiumTiers!.COMPLEX).toEqual(customComplex);
-    expect(result.premiumTiers!.SIMPLE).toEqual(
-      DEFAULT_ROUTING_CONFIG.premiumTiers!.SIMPLE,
-    );
+    expect(result.premiumTiers!.SIMPLE).toEqual(DEFAULT_ROUTING_CONFIG.premiumTiers!.SIMPLE);
   });
 
   it("disables premiumTiers when set to null", () => {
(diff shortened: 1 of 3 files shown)


kagura-agent:
Great analysis @1bcMax — makes total sense that this only fixes half of #148 without the `agenticMode: false` override fix in v0.12.141.

I'll go with **(b)**: rebase onto main, drop the duplicate `mergeRoutingConfig` changes, and keep just the `mergeTierRecord` extraction + the merge test suite as a refactor follow-up. Will update the PR shortly.

kagura-agent:
Closing — v0.12.141+ already includes a comprehensive fix for the tier merging, including the `mergeTiers` helper and `null` handling that this PR aimed to add. Thanks @1bcMax for the quick fix upstream! 🙏

PR #43, opened 2026-02-19
feat: add custom paymentFetch option to proxy

matiasedgeandnode:
## Summary

- Adds optional `paymentFetch` parameter to `startProxy()` for custom payment handlers
- Enables integration of alternative payment systems (spend limits, budget enforcement, etc.)
- Falls back to built-in x402 payment handler when not provided

## Test plan

- [x] Verify existing behavior unchanged when `paymentFetch` is not provided
- [x] Test custom `paymentFetch` function is called when provided

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #166, opened 2026-05-13
docs: add spending policy integration guide (LetAgentPay)

maximberg:
ClawRouter gives agents wallets and x402 micropayments — elegant infrastructure for "how does an agent pay". There's currently no built-in budget cap, so a runaway agent loop can drain its wallet across hundreds of routed calls in seconds. We see ClawRouter users asking about this gap.

This PR adds a single doc — `docs/spending-policies-with-letagentpay.md` — that shows the wrapper pattern for putting a vendor-neutral spending policy engine ([LetAgentPay](https://letagentpay.com), open-source, MIT) in front of ClawRouter calls:

- Pre-flight `requestPurchase` → ClawRouter `chat` → `confirmPurchase` with actual cost
- Working TS snippet
- Starter policy JSON
- Notes on estimate-vs-actual, failed routes, self-host

Also adds a row to the Documentation table in the README.

Disclosure: I work on LetAgentPay. Wrote this because the question keeps coming up and the integration is straightforward. Happy to revise tone, cut promotional language, or move into a different section if you'd prefer a different shape.

— Maxim Berg

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive guide on implementing spending policies with LetAgentPay integration, including pre-flight spending authorization, policy configuration samples, and failure handling strategies.
  * Updated README with link to new spending policies documentation.

[Review Change Stack](https://app.coderabbit.ai/change-stack/BlockRunAI/ClawRouter/pull/166)

### DIFF
--- a/README.md
+++ b/README.md
@@ -561,6 +561,7 @@ Run Claude Code with 50+ models, no rate limits, no Anthropic account, no phone 
 | [Architecture](docs/architecture.md)                   | Technical deep dive      |
 | [Configuration](docs/configuration.md)                 | Environment variables    |
 | [Troubleshooting](docs/troubleshooting.md)             | Common issues            |
+| [Spending Policies with LetAgentPay](docs/spending-policies-with-letagentpay.md) | Cap agent wallet spend   |
 
 ### Blog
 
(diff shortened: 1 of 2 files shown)


VickyXAI:
Thanks @maximberg — closing, but for a reason that only became true after you opened this.

ClawRouter now does spend control natively: `src/spend-control.ts` ships a `SpendControl` class with perRequest / hourly / daily / session windows and file + in-memory storage. So a guide pointing operators at a third-party service for wallet spend caps would send them away from a feature they already have. (We closed #43 for the same reason — its motivation was 'spend limits, budget enforcement', which `spend-control.ts` answers.)

No objection to LetAgentPay as a product. But our docs shouldn't route users to a third party for something built in.

PR #204, opened 2026-07-12
feat(x402): optional TWZRD pre-sign gate (onBeforePaymentCreation)

twzrd-sol:
## Summary

Opt-in buyer-side trust gate on the **same** official x402 client ClawRouter already builds for USDC payments.

When `CLAWROUTER_TWZRD=1`, ClawRouter dynamic-imports optional peer [`twzrd-x402-gate`](https://www.npmjs.com/package/twzrd-x402-gate) and registers `onBeforePaymentCreation` **after** EVM/SVM scheme registration and **before** `createPayFetchWithPreAuth`.

Default ClawRouter path is unchanged (gate is off unless env is set).

### Why lifecycle hook (not probe-then-pay)

```
probe request -> score challenge A -> second request -> may sign challenge B   # TOCTOU
official client -> select requirement -> onBeforePaymentCreation -> sign same  # bound
```

Scores the exact `selectedRequirements` (payTo / network / amount / resource) the client is about to sign.

### Defaults when enabled

| Behavior | Default |
| -------- | ------- |
| Decision-only (`decision=block` aborts) | on |
| Merchant-card wash refuse | on |
| Strict `can_spend=false` block | off (`CLAWROUTER_TWZRD_GATE_ON_CAN_SPEND=1`) |
| Missing peer while env set | log + continue without gate |

### Usage

```bash
npm i twzrd-x402-gate
export CLAWROUTER_TWZRD=1
# optional stricter:
# export CLAWROUTER_TWZRD_GATE_ON_CAN_SPEND=1
```

Docs: `docs/configuration.md` + Payment section of README.

## Test plan

- [x] `npx vitest run src/twzrd-payment-gate.test.ts` (13 tests)
- [x] `npx tsc --noEmit` clean
- [x] Adjacent x402/builder-code tests still pass
- [ ] With gate off: proxy starts and pays as today
- [ ] With `CLAWROUTER_TWZRD=1` + package installed: log shows pre-sign gate ON; wash/block aborts before sign
- [ ] With env set but package missing: warning log, payments continue

## Links

- https://docs.x402.org/advanced-concepts/lifecycle-hooks
- https://intel.twzrd.xyz
- https://www.npmjs.com/package/twzrd-x402-gate (0.5.4+)

## Summary by CodeRabbit

- **New Features**
  - Added an optional TWZRD pre-sign trust gate for x402 payments.
  - When enabled, the gate scores the final selected payment requirements before wallet signing.
[...]

PR #205, opened 2026-07-13
feat(x402): generic beforePayment pre-sign hook (vendor-neutral)

twzrd-sol:
## What

Adds a generic `beforePayment` lifecycle hook to `startProxy`.

Fires after ClawRouter selects a payment requirement, **before** the transaction is built and signed — the last deterministic checkpoint where a payment can be refused without a wallet signature.

```ts
const handle = await startProxy({
  wallet: "0x...",
  beforePayment: async ({ selectedRequirements, resourceUrl }) => {
    if (!ALLOWED.has(selectedRequirements.payTo)) {
      return { abort: true, reason: "payTo not on allowlist" };
    }
    // return void → proceed to signing
  },
});
```

## Design

- **No vendor dependency.** The hook is a plain callback — the caller wires whatever policy they own (allowlist, spend cap, trust preflight, compliance check). No new packages, no new env vars.
- **Official x402 lifecycle.** Uses `x402.onBeforePaymentCreation`, the same hook BlockRun's own builder-code stamp and chain logging already use.
- **Return `{ abort: true, reason }`** to refuse — wallet never signs, payment never reaches the merchant.
- **Return `void` / `{ abort: false }`** to proceed unchanged.
- **Async-safe** — the hook may be async (e.g. to call an external policy service).
- **Fail-closed** — an unhandled throw inside the hook aborts the payment.

## Relationship to the withdrawn #204

#204 asked ClawRouter to add a branded peerDep (`twzrd-x402-gate`) and TWZRD-named env vars. This PR replaces that with a generic, no-dependency seam. The same pre-sign outcome, but the policy is the caller's responsibility — not a specific package.

## Changes (+65 lines, 1 file)

- `src/proxy.ts`: `ProxyOptions.beforePayment` type + `onBeforePaymentCreation` registration in `startProxy`

No test file yet — if the shape is right, I'll add one.

🤖 Generated with [Hermes Agent](https://hermes-agent.nousresearch.com)

## Summary by CodeRabbit

* **New Features**
  * Added an optional pre-payment approval hook for proxy integrations.
  * Payment requests can now be reviewed and approved or refused before transactions are created and signed.
  * Refused payments return a configurable reason or a default refusal message.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -1308,7 +1308,46 @@ export type ProxyOptions = {
    * Example: "socks5://127.0.0.1:1080"
    */
   upstreamProxy?: string;
+  /**
+   * Pre-sign payment hook. Fires after ClawRouter selects a payment
+   * requirement, BEFORE the transaction is built and signed.
+   *
+   * Return `{ abort: true, reason }` to refuse the payment — the wallet
+   * never signs, the request never reaches the merchant. Return `void` or
+   * `{ abort: false }` to proceed.
+   *
+   * No vendor dependency. Wire any policy you own — allowlist, spend cap,
+   * trust preflight, compliance check. The hook receives the exact
+   * selectedRequirements (payTo, network, amount, asset, resource) the
+   * x402 client is about to sign.
+   *
+   * Async-safe: the hook may be async (e.g. to call an external policy
+   * service). An unhandled throw aborts the payment (fail-closed).
+   */
+  beforePayment?: (
+    ctx: BeforePaymentCreationContext,
+  ) => Promise<BeforePaymentCreationResult> | BeforePaymentCreationResult;
 };
+
+/** Context passed to beforePayment hook (subset of @x402/fetch hook context). */
+export interface BeforePaymentCreationContext {
+  selectedRequirements: {
+    scheme: string;
+    network: string;
+    amount: string;
+    asset: string;
+    payTo: string;
+    maxTimeoutSeconds?: number;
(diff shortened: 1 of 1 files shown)


VickyXAI:
Thanks — and credit where it's due: the design is careful. The hook is abort-only (a non-abort return is discarded, so it can't mutate payTo/amount/asset), fail-closed on throw, no new deps. I verified `onBeforePaymentCreation` genuinely exists in @x402/core. This is the only one of the six open payment PRs that still applied to main.

Closing on product grounds rather than code.

**The capability is already native.** `src/spend-control.ts` ships a `SpendControl` class with perRequest / hourly / daily / session windows and file + in-memory storage — that *is* our pre-sign policy mechanism, and it's the reason #43 got closed too. A generic seam adds a second, weaker way to do the same thing.

**And the seam has no first-party consumer.** `startProxy` is library surface the CLI never reaches, so in practice the only thing that would use `beforePayment` is an external gate — which is what #204 asked for directly before it was withdrawn and reframed here as vendor-neutral. We'd rather own policy in `spend-control.ts` than expose a public hook on the money path for one integrator.

One correction for the record: the body says this is "the same hook BlockRun's own builder-code stamp and chain logging already use" — those use `onAfterPaymentCreation` (src/proxy.ts:2019, 2028). The *before* hook is unexercised in this codebase, and the PR ships no tests for it, so it wouldn't have landed as-is regardless.

If `spend-control.ts` is missing a policy you need, an issue describing that gap is very welcome — that we'd absorb.

Issue #202, opened 2026-07-10
GPT-5.6 Sol is not working: upstream 500/server_error


0xCheetah1:
## Summary

`openai/gpt-5.6-sol` is not working through ClawRouter. Requests fail with upstream `server_error` / HTTP 500, and in direct curl testing can hang until the client times out.

## Observed on testbed

- Model: `openai/gpt-5.6-sol`
- Prior request attempt signed Base payments, then failed twice after long waits with upstream `server_error` / HTTP 500.
- Total observed latency for that failed attempt was about 251 seconds.
- Direct curl smoke test against local ClawRouter also did not complete: timed out after 180 seconds with no response body (`HTTP_STATUS:000`).

## Expected

Requests to `openai/gpt-5.6-sol` should complete successfully. If the upstream is broken, ClawRouter should classify/handle the failure instead of leaving users stuck on long failed attempts.

## Notes

For now, `openai/gpt-5.6-terra` appears safer as the generic GPT-5.6 target while Sol is failing.

VickyXAI:
Thanks for the detailed report, @0xCheetah1 — the latency numbers and the "Terra is safer as the generic target" note were spot on.

**Fixed in [v0.12.219](https://github.com/BlockRunAI/ClawRouter/releases/tag/v0.12.219) (now published).** The generic GPT-5.6 shorthands — `gpt5`, `gpt-5.6`, and `openai/gpt-5.6` — now resolve to the stable **Terra** tier instead of the flaky Sol tier, so the default path no longer hits the upstream `server_error` / ~251s hangs you saw. `gpt5` was also bumped off the older `gpt-5.5` generation onto 5.6/Terra.

**Note on explicit Sol pins:** `openai/gpt-5.6-sol` (and the other exact tier pins) stay exact by design — Sol remains reachable for callers who explicitly want the deepest-reasoning tier and accept the risk. So a hard-pinned `gpt-5.6-sol` can still 500/hang while OpenAI's deepest tier is unstable under long-horizon load; that part is upstream. If Sol keeps misbehaving we can revisit adding failover/timeout handling for explicit pins — happy to track that separately if it's still hurting you.

Closing as the generic-alias regression is resolved. Please reopen if the default path still routes you to Sol.

PR #206, opened 2026-07-14
Fix OpenClaw model picker sync

0xCheetah1:
## Summary

- sync OpenClaw provider, allowlist, and model cache order during ClawRouter setup/install
- patch OpenClaw picker ordering from the ClawRouter install path so BlockRun models stay curated in Telegram
- stop advertising duplicate Gemini alias/redirect models while keeping aliases callable
- add Gemini Pro aliases and group curated models by provider
- fix the bundled createRequire helper conflict in generated CLI output

## Verification

- `npm run format:check`
- `npm run typecheck`
- `npm run lint`
- `npm test -- src/models.test.ts src/top-models.test.ts src/index.config.test.ts`
- `npm run build`
- `node dist/cli.js --version`
- testbed install/setup verified with OpenClaw gateway and ClawRouter services active


VickyXAI (changes requested):
Thanks @0xCheetah1 — and credit first, because it's owed: **the tsup banner change buried in this PR was catching a P0.** v0.12.220 was on npm with a CLI that wouldn't start at all (`SyntaxError: Identifier '__cjs_createRequire' has already been declared`), and `@blockrun/mcp` couldn't boot either through `@blockrun/llm`. Nothing in `build && typecheck && test` ever loads `dist/`, so CI published it happily and none of us noticed. That's shipped as v0.12.221–223, and you're credited in the CHANGELOG.

Digging into it turned up the deeper cause, which is worth sharing since it changes what this PR should keep:

**The banner collision was a symptom.** We and `@blockrun/llm` depend on each other, and `noExternal: [/.*/]` inlines it. v0.12.220's Polymarket port added `src/polymarket/fund.ts` — our first-ever `@blockrun/llm` import — so its `import { ... } from "@blockrun/clawrouter"` resolved through `node_modules` to the **last published build of this package**, and esbuild inlined that entire stale bundle beside the one it was building. Two ClawRouters, different versions, separate module state. The old copy carried its own `__cjs_createRequire`, and esbuild can't rename around a banner it never sees (banners are raw text injected *after* bundling) — hence the collision. Renaming fixes the crash but leaves the stale copy: v0.12.221 still shipped at 10.2MB with a whole second ClawRouter inside, and every release would have inlined the previous one (~20MB next). The real fix is an esbuild `alias` pinning that back-import to our own source — **10.2MB → 7.6MB**, on `main` now. So please keep the rename (good defense-in-depth) but drop the expectation that it's the whole story.

## Blocking: the `models.ts` change guts `/v1/models`

This is the one thing that can't land as-is. Running both list-builders over the current head (`86091aaf`) vs `main`:

```
main  (v0.12.223): 205 entries
pr206 (86091aaf) :  83 entries   → 120 aliases lost
```

Everything lost is a friendly alias: `sonnet`, `opus`, `haiku`, `gpt5`, `kimi`, `flash`, `glm`, `codex`, `dalle`, `nano-banana`, … The cause is in `src/models.ts`: `ALIAS_MODELS` is deleted and the catalog filter narrows from `!(m.id in MODEL_ALIASES)` to `!(m.id.includes("/") && m.id in MODEL_ALIASES)`. `OPENCLAW_MODELS` is the sole input to `buildProxyModelList()` (`src/proxy.ts`), whose docstring is explicit:

> This includes alias IDs (e.g., `flash`, `kimi`) so `/model <alias>` works reliably. The picker visibility filter lives at the openclaw.json allowlist layer, not here — `/v1/models` stays open for API-level discovery.

`resolveModelAlias()` still resolves them, but any client that validates against `/v1/models` first — including OpenClaw's `/model` — will reject them. (The other 2 of the 122 are `claude-fable-5` / `grok-4.5`, just branch staleness; a rebase picks those up.)

**The good news: you don't need this change to get the dedup.** Which brings me to the part of this PR that was dead right.

## You found a real bug, and the cache sync was the correct fix

The duplicate/stale rows come from a **third** model-list plane — `~/.openclaw/agents/<agent>/agent/models.json` — separate from `openclaw.json`'s picker and allowlist. Nothing ever synced it. Measured on a real machine *after* `clawrouter setup` had already repaired `openclaw.json` to the correct 47, that cache still held **155 entries — 127 retired** (`gpt-5.2`, `gpt-4.1`, `o1`, …), duplicate `free` / `moonshot/kimi-k2.5` rows, and none of the current flagships.

Your model-cache sync targets exactly that. It's shipped in **v0.12.223** as `syncAgentModelCache()`, generalized a little: iterates all agent dirs rather than hardcoding `main`, only repairs an existing `blockrun` provider (never introduces one), preserves other providers and `baseUrl`/`api`/`apiKey`, compares order positionally, and is gated outside gateway mode. Your positional-comparison insight for the picker/allowlist is in there too. Thank you — that diagnosis is what made it findable.

## Two things I'd drop

**Patching OpenClaw's `dist/*.js`.** I'm fairly sure I know why you went there — `openclaw plugins install` ends in `Config write rejected: openclaw.json (size-drop:76642->25200)`. OpenClaw guards against large config shrinks, and pruning 155 stale models → 47 *is* a large shrink, so it rejects the write and rolls the install back, restoring the stale list. It's a genuine wall and you were right that something is broken. But string-replacing another package's shipped bundle gets reverted by any `openclaw` update, and writing OpenClaw's own state from under its install transaction is what stranded users on the 2026.5.2 baseHash rollback. `clawrouter setup` sidesteps it (it writes directly) and is the supported repair path. I think the real fix belongs upstream in OpenClaw — happy to co-write that issue with you.

FWIW I tested the obvious workaround (prune first, then install) and it does **not** work: OpenClaw compares against a cached baseline, not current disk.

**`postinstall` auto-running `setup`.** It runs on every `npm install` — CI, Docker, a dev's local `npm install` — and silently mutates the user's global `~/.openclaw` config. If it ever fires inside OpenClaw's own plugin-install transaction it's the baseHash rollback again. The discoverability problem it solves is real (a bare `npm install -g` does zero integration); I'd rather solve it with a loud hint than an implicit global write.

Also worth knowing, since it bit this exact flow: **anything loose under `scripts/` that imports `child_process` makes the plugin uninstallable** — OpenClaw's scanner blocks the whole install. I shipped that bug myself in v0.12.222 and fixed it in v0.12.223; `dist/` is exempt.

## Suggested path

Rebase on `main` (v0.12.223) and keep: the openclaw detection candidates in `cmdSetup`, the Gemini Pro aliases, and the positional comparisons. Drop: the `models.ts` list rewrite, the OpenClaw dist patch, and the postinstall. The tsup and cache-sync pieces are already in.

On `top-models.json`: the `gpt-5.6-sol`-before-`terra` reorder matches blockrun's registry order, but it re-fronts the tier that 500s after ~250s waits (#202) and that v0.12.219 deliberately routed generic aliases away from. Fine either way — just flagging it was intentional.

Really good catch on the P0. Happy to take the rebase in pieces if that's easier.

VickyXAI:
Closing this — but not because it was wrong. Almost everything in it either shipped or led directly to something that did, and the events of the last few hours have overtaken the branch to the point where rebasing it would cost you more than it's worth. Here's the full accounting, because you're owed it.

## What you caught

**The tsup banner change was catching a P0.** v0.12.220 was on npm with a CLI that would not start (`SyntaxError: Identifier '__cjs_createRequire' has already been declared`), and `@blockrun/mcp` couldn't boot either through `@blockrun/llm`. Nothing in `build && typecheck && test` loads `dist/`, so CI published it and nobody noticed. You found it and buried it under "fix the bundled createRequire helper conflict" — it was the most important thing in this PR by a wide margin. Shipped in **v0.12.221**, credited in the CHANGELOG.

Chasing *why* it happened is what produced the rest:

- **v0.12.222** — the collision was a symptom. `noExternal` was inlining an entire stale published ClawRouter beside the one being built (10.2MB → 7.6MB once aliased). The old copy carried its own banner, and esbuild can't rename around text injected after bundling.
- **v0.12.225** — the actual root cause: we were pinned to `@blockrun/llm ^2.11.0`, and 2.13.0 lists `@blockrun/clawrouter` in its `dependencies`. 3.x demotes it to a peer, so npm no longer installs a stale copy of us into our own tree. **Your one-line rename led to a dependency cycle we'd been shipping for months.**

## What shipped from this PR

- **Model cache sync → v0.12.223** as `syncAgentModelCache()`. You were right that `~/.openclaw/agents/<agent>/agent/models.json` is a third, unsynced plane. Measured on a real box *after* `setup` had repaired `openclaw.json` to 47: that cache still held **155 entries, 127 retired**, with duplicate `free`/`kimi-k2.5` rows. Generalized a little — iterates all agent dirs instead of hardcoding `main`, only repairs an existing `blockrun` provider, preserves other providers and `baseUrl`/`api`/`apiKey`, positional order compare, gated outside gateway mode.
- **Positional comparison** for the picker/allowlist — same release.
- **Allowlist preservation** → merged as **#207**. (You'd already split it out and reverted it here.)
- **Banner rename** → v0.12.221, kept as defense-in-depth even after the alias made it redundant.

## Why closing instead of asking you to rebase

The branch is `CONFLICTING` now — main moved through v0.12.221→226 and touched `models.ts`, `top-models.json`, `tsup.config.ts`, `package.json`, `dist`. Rebasing 500 lines to land what's left wasn't a good trade for your time, so **I've cherry-picked the two remaining pieces into v0.12.226** with attribution:

1. **The openclaw detection candidates in `cmdSetup`** — you were right, and it's easy to demonstrate: under `env -i PATH=/usr/bin:/bin`, `command -v openclaw` finds nothing while the binary sits right there in the nvm prefix. npm runs lifecycle scripts exactly like that, so setup would tell users to install what they already had. Your fallback chain (`npm_config_prefix` → `npm root -g` → `~/.npm-global/bin` → `~/.local/bin` → `/usr/local/bin`) fixes it.
2. **`gemini-pro` / `gemini-3-pro`** (plus `gemini-3.1-pro`) → `google/gemini-3.1-pro`. `gemini-3-pro` was never a real id, but people reach for it and got a 400.

One detail worth knowing for next time, since it's subtle and you had it right by accident: those had to be **bare** alias keys. A slash-prefixed key like `google/gemini-3.1-pro` would have shadowed the catalog entry — `OPENCLAW_MODELS` filters out any catalog model whose id is an alias key — and the row would then carry the alias's metadata instead of the model's real pricing. There's now a test pinning that.

The `models.ts` change is the one thing I'd ask you not to carry forward: it drops 120 aliases from `/v1/models` (205 → 83 measured against your head), and `buildProxyModelList()` is the sole consumer, so `/model sonnet` and friends break for anything that validates against the list first. You never needed it for the dedup — the cache sync was the fix.

## The wall you hit

For the record, since it's the one thing here I'd push back on and I want to be straight about why: the OpenClaw `dist/*.js` patching. You hit `Config write rejected: openclaw.json (size-drop:76642->25200)` — OpenClaw guards against config shrinks >50%, and pruning 155 stale models to 47 is one. That's a real wall and you were right that something is broken.

It turns out upstream already has an issue open for exactly this — [openclaw/openclaw#82124](https://github.com/openclaw/openclaw/issues/82124), whose own example is `models.providers.<provider-id>`. I've [added our production repro](https://github.com/openclaw/openclaw/issues/82124#issuecomment-4976213573) with the numbers, including one thing that'll save you time: **pre-shrinking doesn't work around it** — the guard compares against a restored snapshot, not current disk, so "prune first, then install" still fails. `clawrouter setup` is unaffected because it writes directly.

One more, learned the hard way on my side: anything loose under `scripts/` that imports `child_process` makes the plugin uninstallable — OpenClaw's scanner blocks the whole install. I shipped that bug myself in v0.12.222 (my own smoke script) and fixed it in v0.12.223. Your `postinstall.mjs` would have tripped it too.

Genuinely good catch on the P0 — it was doing real damage and none of us had noticed. Thanks.

Issue #213, opened 2026-07-18
Kimi K3 emits OpenClaw tool calls as text instead of structured tool_calls


0xCheetah1:
## Summary
Kimi K3 is not reliably using OpenClaw tool calling through ClawRouter. On the Blockrun Testbed Telegram/OpenClaw setup, it emits tool-call-looking JSON or prose as assistant text instead of structured `tool_calls`, so users see raw JSON/instructions rather than the tool executing.

## Examples
Prompt:

```
what is in this dir, and can you help me check when argentina are playing the finals of the world cup and who they are playing it against?
```

Observed assistant text:

```
Let's do web_search.
{"query": "2026 FIFA World Cup final date Argentina", "top_n": 5, "recency_days": -1}
```

Prompt:

```
what's in the identity md file
```

Observed assistant text:

```
{"path":"/home/Blockrun/.openclaw/workspace/IDENTITY.md","action":"read"}
```

Earlier related observed shape:

```
{"cmd":["bash","-lc","ls -1a"], "timeout": 10000}
```

## Expected
Kimi K3 should emit structured OpenAI-compatible `tool_calls` so OpenClaw executes the intended tool and suppresses raw assistant tool-call text.

## Actual
Tool intent is leaked as plain assistant content. The tool is not executed, and the user sees raw JSON/prose.

## Environment
- OpenClaw gateway on Blockrun Testbed
- ClawRouter OpenClaw plugin path
- Model selected via `/model`: `blockrun/moonshot/kimi-k3`

## Notes
This likely needs investigation at the OpenClaw prompt/schema compatibility layer and/or ClawRouter textual tool-call recovery. The current behavior has multiple Kimi K3 leak shapes, so a one-off parser patch is probably insufficient without a more systematic compatibility fix.

VickyXAI:
Fixed in **v0.12.230** (published to npm).

**Root cause:** ClawRouter already synthesizes structured `tool_calls` from four textual leak shapes, but each needs the tool *name* inside the text. Kimi K3 instead emits the tool's *arguments* as a bare JSON object with no `name`/`type` field — the name is only in the preceding prose (`Let's do web_search.\n{...}`) or absent entirely (`{"path":...,"action":"read"}`, `{"cmd":[...],"timeout":...}`) — so none of the existing extractors fired.

[...]

Issue #193, opened 2026-06-29
GPT 5.4 emits tool calls as plain text


0xCheetah1:
## Problem

GPT 5.4 sometimes returns tool calls as plain assistant text instead of structured OpenAI-compatible `tool_calls`.

Observed in Chermes/Telegram:

```text
{
  "type": "function",
  "name": "terminal",
  "parameters": {
    "cmd": "ls -alh /home/Blockrun"
  }
}
```

Also observed:

```text
terminal
ls -alh /home/Blockrun
[/terminal]
```

and earlier during the same investigation:

```text
{"name":"session_search","parameters":{"query":"\"I though he never left\" \"Ronaldo\""}}
```

```text
read_file(parameters={"path":"/home/Blockrun"})
```

## Expected

GPT tool calls should reach downstream clients as structured OpenAI-style `tool_calls`, with empty assistant content and `finish_reason: "tool_calls"`.

## Actual

The model sometimes emits JSON/function-call-looking text in `message.content`. Downstream chat surfaces then display that raw text to the user instead of dispatching the tool.

## Notes

This is similar to #189 / #190 for Gemini 3.5 Flash, but with GPT-specific text shapes instead of `[Called function "NAME" with args: {...}]`.

A local testbed patch following the same approach as #190 recovered these shapes in `extractTextualToolCalls`:

- standalone `{"name":"NAME","parameters":{...}}`
- standalone `{"type":"function","name":"NAME","parameters":{...}}`
- whole-content `NAME(parameters={...})`
- trailing JSON object after prose only when `type` is exactly `function`
- whole-content `terminal\nCOMMAND\n[/terminal]`

Guardrails in the local patch:

- no prose JSON examples unless the whole content is a call
- no trailing prose JSON unless `type:function` is explicit
- no incomplete terminal block
- terminal `cmd` normalized to `command` while preserving `cmd`

Targeted tests on the patched testbed checkout passed:

```text
41 passed
```

Checks also passed on the patched testbed checkout:

```text
npm run typecheck
npm run lint
npm run format:check
npm run build
```

houleixx:
This is exactly the kind of issue where request/response visibility helps.

ccglass can show the outgoing `tools` / `tool_choice` request and the raw assistant response, including whether the provider returns real `tool_calls` or plain assistant text.

Project: https://github.com/jianshuo/ccglass

That could help separate model behavior from router/client parsing behavior and produce a clearer repro.

[...]

Issue #189, opened 2026-06-28
Gemini 3.5 Flash emits tool calls as text transcripts


0xCheetah1:
## Problem
Gemini 3.5 Flash sometimes returns tool calls as plain assistant text instead of structured OpenAI-compatible `tool_calls`.

Observed in Chermes/Telegram:

```text
[Called function "terminal" with args: {"command":"whoami"}]
```

and:

```text
[Called function "search_files" with args: {"pattern":"*","target":"files"}]
```

Other models generally return proper structured tool calls, so this appears specific to Gemini 3.5 Flash behavior through the OpenAI-compatible path.

## Expected
Gemini tool calls should reach downstream clients as structured OpenAI-style `tool_calls`, with empty assistant content and `finish_reason: "tool_calls"`.

## Actual
The model sometimes emits a text transcript, which downstream chat surfaces may display to the user instead of dispatching the tool.

## Testbed notes
A local testbed patch that parses this shape:

```text
[Called function "NAME" with args: {...}]
```

and converts it into a structured tool call made direct ClawRouter requests work for both non-streaming and streaming responses.

Example direct streaming response after the patch included:

```json
{
  "delta": {
    "tool_calls": [
      {
        "type": "function",
        "function": {
          "name": "terminal",
          "arguments": "{\"command\":\"whoami\"}"
        }
      }
    ]
  }
}
```

Targeted tests added on testbed passed:

```text
27 passed
```

Checks also passed on the patched testbed checkout:

```text
npm run typecheck
npm run lint
npm run build
npm run format:check
```

## Related symptom
After adding runtime normalization, Hermes/Telegram can execute the tool, but may still display raw tool result lines like:

```text
[Tool "function" returned]: {...}
```

That may be a separate display-layer issue, but the original failure is Gemini returning tool calls as text transcripts instead of structured calls.

PR #215, opened 2026-07-20
feat(models): add Qwen3.7 Max

KillerQueen-Z:
## Summary

- Add Qwen3.7 Max to the ClawRouter catalog and curated top-model list.
- Add `qwen`, `qwen3.7-max`, and `qwen-3.7-max` aliases.
- Preserve gateway pricing at raw $1.475/$4.425 per million tokens; settlement applies BlockRun's standard margin.

## Validation

- `npm test -- --run src/models.test.ts src/top-models.test.ts src/proxy.models-endpoint.test.ts`
- `npm run typecheck`

### DIFF
--- a/src/models.test.ts
+++ b/src/models.test.ts
@@ -70,6 +70,22 @@ describe("resolveModelAlias", () => {
     expect(BLOCKRUN_MODELS.some((m) => m.id === "anthropic/claude-sonnet-4.5")).toBe(true);
     expect(resolveModelAlias("sonnet")).toBe("anthropic/claude-sonnet-4.6");
   });
+
+  it("registers Qwen3.7 Max with its public aliases and gateway pricing", () => {
+    const qwen = BLOCKRUN_MODELS.find((model) => model.id === "qwen/qwen3.7-max");
+
+    expect(qwen).toMatchObject({
+      name: "Qwen3.7 Max",
+      inputPrice: 1.475,
+      outputPrice: 4.425,
+      contextWindow: 1_000_000,
+      maxOutput: 65_536,
+      reasoning: true,
+      toolCalling: true,
+    });
+    expect(resolveModelAlias("qwen")).toBe("qwen/qwen3.7-max");
+    expect(resolveModelAlias("qwen3.7-max")).toBe("qwen/qwen3.7-max");
+  });
 });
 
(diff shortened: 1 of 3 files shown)


VickyXAI:
Thanks for this @KillerQueen-Z — the model data was correct, and it shipped in **[v0.12.231](https://github.com/BlockRunAI/ClawRouter/releases/tag/v0.12.231)**.

Your commit is on `main` as [999d6aa](https://github.com/BlockRunAI/ClawRouter/commit/999d6aa) with your authorship intact, and you're credited in the CHANGELOG and the release notes.

I landed it via a separate branch rather than merging here, because this branch forked at v0.12.216 and `src/top-models.json` had moved a lot since — GitHub reports it as `CONFLICTING`, and both of the obvious resolutions are wrong:

- **take-theirs / union** resurrects four models that were retired upstream (`moonshot/kimi-k2.7`, `xai/grok-3`, `xai/grok-4-0709`, `free/gpt-oss-120b`) and creates literal duplicates of four live ones
- **take-ours** silently drops `qwen/qwen3.7-max` — the model would ship registered but invisible in the picker

So it needed a rebase, not a conflict resolution. I also folded in a few things while it was open:

**Dropped the bare `qwen` alias.** Every other `qwen*` shorthand in the table resolves to a **free** model (`qwen-coder`, `qwen-thinking`, `qwen3-next`, `qwen3.5-122b`), so binding the shortest name to a $1.475/$4.425 flagship would silently bill people who typed it expecting the free tier. Same reasoning that keeps generic `kimi` on K2.7. The explicit pins stay, and I added `qwen3-7-max` for parity with the `sonnet-4.6`/`sonnet-4-6` convention.

**Verified `toolCalling: true` against the live gateway** rather than assuming it — it does return a structured `tool_calls` array (valid JSON arguments, `finish_reason: "tool_calls"`), no textual leak of the kind we had to work around for Kimi K3 (#213), Gemini (#189) and GPT (#193). Good call flagging it true; now it's pinned with evidence.

**Filled in the rest of the model-add checklist:** README pricing table, `skills/clawrouter/SKILL.md` catalog line, CHANGELOG + version bump. Worth knowing for next time — a model add touches five files, not three.

Closing this as landed. Genuinely useful contribution, and the pricing matched blockrun's catalog exactly — please send more.

PR #218, opened 2026-07-25
feat(proxy): default-on TWZRD AutoGate pre-spend check before x402 sign

twzrd-sol:
## Why

ClawRouter is a high-traffic agent pay path (local proxy signs x402 for BlockRun models / Surf / etc.). Outer `wrapFetch` never sees the 402 — payment is internal. A third-party pre-spend check only works if it runs **on the internal `x402Client` before sign**.

TWZRD AutoGate (`installTwzrdAutoGate`) is that seat: free Solana wash/preflight, observe-mode on Base (no fake scores), kill switch env, measurable `X-Twzrd-Caller` identity.

## What

After `new x402Client()` + scheme registration in `proxy.ts`:

```ts
await installTwzrdAutoGateOnClient(x402);
```

- **Default ON** via dependency `twzrd-x402-gate@^0.8.4`
- **Kill:** `TWZRD_AUTO_GATE=0` or `TWZRD_GATE_ENABLED=false`
- **Base/EVM:** `unsupportedNetworkMode: \"observe\"` — allow; Solana corpus does not invent Base reputation
- **Solana:** full preflight; `refuseWashFlagged: true`; `gateOnCanSpend: false` (decision=block still aborts)
- Soft-fail if package missing at runtime (proxy still boots)

## Docs

README section **Pre-spend trust (TWZRD AutoGate)**.

## Tests

`src/twzrd-autogate.test.ts` — 3 passed (env kill + install surface).

## Out of scope

- Changing BlockRun facilitator or routing
- Scoring Base merchants as Solana wallets
- Facilitator volume / marketplace features

## Fork-1 note (TWZRD)

This is a distribution pin: default check on a hot agent pay path, opt-out documented. Not a facilitator PR.

## Summary by CodeRabbit

* **New Features**
  * Added an optional **TWZRD AutoGate** pre-spend guard that runs before payment signing, with network-specific behavior (including observe mode on EVM and abort behavior on Solana when flagged).
  * Proxy startup now conditionally installs the guard when available and continues with the existing payment flow.
  * Added environment-variable kill switches to disable the gate.

* **Documentation**
  * Updated the README with a new **“Pre-spend trust (TWZRD AutoGate)”** section describing behavior, activity indicators, and how it complements routing.

* **Tests**
  * Added a Vitest suite covering disablement via environment variables and successful guard installation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,25 @@ ClawRouter is built for the agent-first world:
 
 This is the stack that lets agents operate autonomously: **x402 + USDC + local routing**.
 
+### Pre-spend trust (TWZRD AutoGate)
+
+By default ClawRouter installs [`twzrd-x402-gate`](https://www.npmjs.com/package/twzrd-x402-gate) on its internal x402 client **before any payment is signed**:
+
+- **Solana** pays: free TWZRD preflight (wash / allow|warn|block). `block` aborts the sign.
+- **Base / EVM** pays: observe mode (no fabricated Solana reputation; payment proceeds).
+- **Seat identity:** every gate call stamps `X-Twzrd-Caller: twzrd-x402-gate/<ver>` for measurable installs.
+
+**Kill switch (default remains ON):**
+
+```bash
+export TWZRD_AUTO_GATE=0
+# or
+export TWZRD_GATE_ENABLED=false
+```
+
+TWZRD is a pre-spend check, not a facilitator and not a replacement for ClawRouter routing.
+
+
 ---
 
 ## How it compares
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,6 +23,7 @@
         "@x402/svm": "^2.9.0",
         "axios": "^1.18.1",
         "https-proxy-agent": "^9.1.0",
+        "twzrd-x402-gate": "^0.8.4",
         "undici": "^8.5.0",
         "viem": "^2.39.3"
       },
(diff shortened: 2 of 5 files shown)


twzrd-sol:
Following up with validation on the current head (`c0852f07`) and a concrete ask, since this has been open a few days.

**The CodeRabbit Major is fixed.** The original `installTwzrdAutoGateOnClient` caught everything, so a hook-registration failure would have let the proxy boot with payments unguarded — the opposite of the PR's purpose. Now only a genuinely missing module soft-skips (`ERR_MODULE_NOT_FOUND` / `MODULE_NOT_FOUND` / "Cannot find module"); every other failure rethrows, so it fails closed. CodeRabbit's review on the current head is passing.

**Cold-clone validation, run from scratch:**
- `npm ci` clean on **Node 22.22.2** (matching `engines: >=22`) and also on Node 20 — 720 packages, no `EUSAGE`, no `@emnapi` resolution churn.
- `npx tsc --noEmit` clean.
- `npx vitest run` → **706 passed / 3 skipped, 61 files**.
- The gate is an `optionalDependency` pinned in the lockfile, so forks that omit `twzrd-x402-gate` are unaffected and still install cleanly.

**The ask — any of these three is a good outcome, and I'd rather have a no than silence:**
1. Merge it as-is (default-on with `TWZRD_AUTO_GATE=0` as the kill switch), or
2. Tell me default-on is the blocker and I'll flip it to opt-in (off unless explicitly enabled) — that's a small change and I'm happy to make it, or
3. Tell me this isn't a direction you want in ClawRouter and I'll close it myself, no hard feelings.

If it helps the decision: the change is dependency-optional, fails closed on real errors, fails *open* only when the package is absent, and adds nothing to the request path when disabled.

twzrd-sol:
Withdrawing this. You already chose the direction in #205 (spend-control.ts); this PR is more vendor-specific and is no longer the right fit. Thanks for the review path — no action needed on your side.

PR #227, opened 2026-07-30
fix(music): default to music-3.0 and list current models

octo-patch:
Reason: refresh the MiniMax music provider to the current music-3.0 model and model list.

The runtime MiniMax music generation provider still defaulted to `minimax/music-2.5+` and advertised only the legacy `music-2.5` variants. This refreshes the registered default and model list to the current generation models.

Changes:
- `src/index.ts` (`buildMusicGenerationProvider`): default model is now `minimax/music-3.0` and the advertised `models` list now includes `minimax/music-3.0`, `minimax/music-2.6`, `minimax/music-3.0-free`, and `minimax/music-2.6-free`.
- `src/proxy.ts` (`/v1/audio/generations` handler): the fallback `audioModel` default is now `minimax/music-3.0`.
- `test/test-e2e.ts`: updated the mock upstream response and the local/live music fixtures to use `minimax/music-3.0`.

Checks run:
- `npx tsc --noEmit` — passed.
- `npx vitest run` — 723 passed / 3 skipped (one flaky lifecycle timeout passed on re-run; unrelated to this change).
- `npx tsx test/test-e2e.ts` — 20 passed / 0 failed (mocked audio generation path exercised with the refreshed model).

## Summary by CodeRabbit

* **New Features**
  * Updated the default music generation model to `minimax/music-3.0`.
  * Added support for `minimax/music-3.0`, `minimax/music-2.6`, and their free variants.

* **Bug Fixes**
  * Audio generation requests without a specified model now use the updated default model.

* **Tests**
  * Updated music generation test coverage to use the latest default model.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1261,8 +1261,8 @@ function buildMusicGenerationProvider(): MusicGenerationProviderPlugin {
   return {
     id: "blockrun",
     label: "BlockRun",
-    defaultModel: "minimax/music-2.5+",
-    models: ["minimax/music-2.5+", "minimax/music-2.5"],
+    defaultModel: "minimax/music-3.0",
+    models: ["minimax/music-3.0", "minimax/music-2.6", "minimax/music-3.0-free", "minimax/music-2.6-free"],
     capabilities: {
       maxTracks: 1,
       maxDurationSeconds: 240,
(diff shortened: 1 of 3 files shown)


VickyXAI:
Closing: this would break music generation. ClawRouter's upstream is BlockRun's gateway, and blockrun `origin/main` has no `minimax/music-3.0`, `music-2.6`, or free variants — the only registered audio model is `minimax/music-2.5+` (`music-2.5` redirects to it). The gateway validates model ids in `/v1/audio/generations` and returns `400 Unknown audio model` for anything else, so this default would fail every music request.

The e2e suite passed because the mock upstream accepts any model string — it can't catch an invalid gateway id. If MiniMax's music-3.0 line lands in blockrun's catalog, we'll sync ClawRouter then (blockrun is source of truth for model ids).

PR #238, opened 2026-08-07
feat(router): make portfolio routing the Auto default

KillerQueen-Z:
## Summary

- make the deterministic Router v3.4 portfolio strategy the default for Auto requests
- classify task shape locally, enforce tool, vision, structured-output, and context constraints, then rank an ordered fallback portfolio
- extract the product-neutral implementation into [`BlockRunAI/router-core`](https://github.com/BlockRunAI/router-core) and pin this product to immutable commit `6a790eb`
- keep `@blockrun/clawrouter/router` as a backwards-compatible re-export for existing SDK consumers
- inject ClawRouter's live model catalog capabilities into Router Core so the product catalog remains authoritative
- retain `strategy: "rules"` as an immediate rollback and optional local-only shadow comparison with no second model call
- bump the release version to `0.12.241`

## Why

The previous Auto path used a fixed tier primary after rule classification. The evaluated portfolio policy improves model selection for agent workflows while keeping routing local and deterministic. In the frozen 100-task, three-arm Franklin evaluation, the new policy scored 57% versus 49% for the previous router and reduced normalized cost per successful task by 6.44%.

Extracting the selector is also a cleaner ownership boundary: Router Core owns classification, hard constraints, portfolio scoring, and fallback ordering; ClawRouter owns its gateway, wallet, payment, proxy, session, and live model catalog. The benchmark remains a separate consumer and is not shipped in either product.

## Safety and rollout

- no benchmark settlement or grading code is included in production
- no prompt is persisted by shadow mode
- shadow mode never issues an additional paid completion
- operators can roll back with `routing.strategy = "rules"`
- `tool_choice: "none"` is authoritative, and host tool descriptions do not create false per-turn tool requirements
- malformed performance telemetry and empty candidate overrides fail safely
- explicit user pins stay sticky only while they satisfy hard context, exclusion, tool, and vision constraints
- the Router Core dependency is an immutable, checksummed HTTPS tarball; it has zero runtime dependencies

## Validation

- Router Core: typecheck, build, package dry-run, and **94 / 94 tests passed**
- ClawRouter: typecheck and production build passed, including dist import smoke checks
- ClawRouter product suites: **713 passed, 3 skipped**
- packed/build artifact smoke suite: **62 / 62 passed**
- real `blockrun/auto` Proxy request completed end to end; the existing insufficient-balance path correctly fell back to the configured free model
- latest `main` model-lifecycle update is synchronized: the retired `seed-oss-36b` fallback was removed once in Router Core and inherited by both products
- Franklin consumer validation against the independent package: **646 / 646 tests passed** and **5 / 5 real Auto CLI E2E paths passed**
- 340k-character Router microbenchmark: 0.125 ms mean over 200 warm iterations on the local test machine

## Summary by CodeRabbit

* **New Features**
  * Added portfolio-based routing that considers task type, tools, vision, context limits, pricing, performance, and model affinity.
  * Added a public router package export and additional routing utilities and types.
  * Added optional shadow-routing telemetry without affecting the serving request.
  * Added model performance profiles and configurable routing strategies.

* **Bug Fixes**
  * Improved fallback handling for capacity, pinned models, tools, compliance, and vision requirements.
  * Added clear errors when no eligible model can handle a request.

* **Tests**
  * Expanded coverage for routing, fallback behavior, tool detection, capacity filtering, and shadow routing.


VickyXAI:
Reviewed and pushed three fixes to this branch (`46d6e9c`). The router extraction and the portfolio policy are unchanged — all three were mechanical.

### 1. The committed `dist/` was stale relative to your pin

`dist/router/index.js` carried `"free/seed-oss-36b"` in three fallback chains (3013 / 3131 / 3221), with the pre-fix comments intact. That model went **HTTP 410 EOL at NVIDIA on 2026-08-03** and v0.12.241 removed it.

Credit where it's due: **your pin was already correct.** router-core `6a790eb` ("fix: retire unavailable seed fallback") fixes all three sites in `config.ts`, and router-core's own `dist/` is clean. Only ClawRouter's committed artifact was built against an older snapshot — `c714217` refreshed it, but from a pre-fix install. Rebuilt.

Why it was blocking rather than cosmetic: BlockRun **server-redirects** retired free ids, so routing to one silently defeats `/exclude`. The caller excludes a model, the router hands it the request anyway, and the gateway answers from the redirect target.

### 2. `@blockrun/router-core` moved to `devDependencies`

tsup's `noExternal` inlines it — `dist/router/index.js` line 3 is `// node_modules/@blockrun/router-core/dist/index.js`, and nothing imports it at runtime. Sitting in `dependencies`, it made every `npm install @blockrun/clawrouter` fetch a `codeload.github.com` tarball at install time for code already in the bundle.

The integrity hash does protect `npm ci` in this repo. But the published package doesn't ship our lockfile, so consumers resolve the bare URL straight from `package.json` with no verification — and installs break whenever GitHub is unreachable or the org is renamed, which has happened here before. The packed tarball now declares zero non-registry dependencies again.

### 3. Version `0.12.241` → `0.12.242`

`0.12.241` is already on npm (it's the seed-oss fix). `publish.yml` guards on `pkg != npm`, so merging at `0.12.241` would land on main and **silently never publish**.

### Added: `src/router/free-model-liveness.test.ts`

Router selection coverage left this repo with the code — measured **713 tests here vs 731 on main**, with `selector.test.ts` and `strategy.test.ts` deleted and none added. Item 1 is exactly what walked through that gap. The 94 tests in router-core are good, but they can't catch a stale build on this side.

The guard walks every tier container's `primary` + `fallback` and asserts each `free/*` id is live in `src/top-models.json`, plus the two deliberate gpt-oss defaults. **The allowed set is derived from the picker, not hard-coded**, so the next free-tier resync updates it for free.

Verified to *fail* against the regression — re-injecting seed-oss into one chain yields `tiers.SIMPLE.fallback[3] -> free/seed-oss-36b` and fails two of three assertions. A third assertion fails loudly if the config shape changes and the walk stops finding tiers, so it can't pass vacuously.

### Verification

typecheck, lint, prettier, build + dist smoke all clean. **716 tests pass** (713 + 3 new). Packed artifact confirmed to declare no URL dependency; `dist/` confirmed free of seed-oss *routing* entries (the four remaining mentions in `dist/index.js` are the deliberate pins and catalog entry that v0.12.241 keeps routable, since the gateway redirects them).

### One thing worth deciding separately

Model lifecycle now has two owners. Every upstream EOL — blockrun re-probes roughly weekly — needs a change in both repos, and the first drift happened within four days. The guard test above is the cheap hedge; a tighter coupling is a real design question, not a blocker for this PR.

Nice work on the extraction. `smoke-dist.mjs` being *extended* to cover the new entry point, rather than loosened around it, is the right instinct.

PR #253, opened 2026-08-18
fix: add client abort to img2img + preserve tool pairs in truncateMessages

### LINKED ISSUE #251: img2img endpoint missing client abort controller charges user on disconnect
Bug

The /v1/images/image2image handler (proxy.ts ~L2698) does not create a clientAbort AbortController, unlike the /v1/images/generations handler (L2498) which correctly cancels upstream requests when the client disconnects.

If a client disconnects mid-request, the upstream call continues and the x402 payment settles the user is charged for a result nobody receives.

Steps to reproduce
Send a request to /v1/images/image2image
Disconnect the client before the upstream response arrives
Observe that the upstream request completes and x402 payment settles despite no client to receive the result
Expected behavior

The upstream request and x402 payment should be cancelled when the client disconnects, matching the behavior of /v1/images/generations.

Suggested fix

Add the same clientAbort pattern already used in the image generation handler:
const clientAbort = new AbortController();
res.on("close", () => {
  if (!res.writableEnded) clientAbort.abort();
});

const upstream = await payFetch(`${apiBase}/v1/images/image2image`, {
  method: "POST",
  headers: { "content-type": "application/json", "user-agent": USER_AGENT },
  body: reqBody,
  signal: clientAbort.signal,  // ← missing today
});

### LINKED ISSUE #252: truncateMessages() breaks tool_call / tool_result pairs at slice boundary
Bug

truncateMessages() (proxy.ts ~L1152) slices conversation messages with conversationMsgs.slice(-maxConversation). This can split an assistant message containing tool_calls from its corresponding tool role response, causing downstream API 400 errors:

Anthropic: tool_use block without matching tool_result
OpenAI: tool_calls referenced but tool response missing
Steps to reproduce
Build a conversation with 200+ messages where a tool_calls assistant message and its tool response straddle the truncation boundary
Send it through ClawRouter
Observe 400 error from the upstream provider
Expected behavior

Truncation should never split a tool_call/tool_result pair. If one side of the pair falls within the kept window, the other side should be included as well.

Suggested fix

After slicing, scan the boundary:

If the first kept message has role: "tool", walk backwards to include its parent assistant message
If the last dropped message has tool_calls, include its corresponding tool responses

Sertug17:
## Describe your changes

Closes #251. Closes #252.

The `/v1/images/image2image` handler was missing a `clientAbort` AbortController, unlike the `/v1/images/generations` handler which correctly cancels upstream requests on client disconnect. This caused unnecessary x402 payment settlement when the client disconnected mid-request charging the user for a result nobody receives.

Additionally, `truncateMessages()` used a naive `slice(-maxConversation)` that could split `tool_call` / `tool_result` message pairs at the truncation boundary. When the slice cut between an assistant `tool_calls` message and its corresponding `tool` response, downstream providers (Anthropic, OpenAI) returned 400 errors due to orphaned tool references.

**Changes:**
- `img2img`: Added `clientAbort` AbortController with `res.on("close")` listener, matching the existing pattern in the image generation handler. Passes `signal: clientAbort.signal` to `payFetch()`.
- `truncateMessages()`: After computing the slice start index, walks backwards past any `tool` role messages at the boundary to keep tool_call/tool_result pairs intact.

## Checklist before requesting a review
- [x] Relevant issues are linked in the PR description (#251, #252)
- [x] Documentation/comments updated according to changes

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved message truncation to preserve assistant tool-call context and prevent incomplete tool interactions.
  * Image-to-image requests now stop source-image downloads, request processing, payment processing, and upstream operations when the client disconnects.
  * Paid API requests now cancel upstream operations when the client disconnects, reducing unnecessary processing and resource use.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -1165,7 +1165,28 @@ function truncateMessages<T extends { role: string }>(messages: T[]): Truncation
 
   // Keep all system messages + most recent conversation messages
   const maxConversation = MAX_MESSAGES - systemMsgs.length;
-  const truncatedConversation = conversationMsgs.slice(-maxConversation);
+  let sliceStart = conversationMsgs.length - maxConversation;
+
+  // Avoid splitting tool_call / tool_result pairs at the slice boundary.
+  // If the first kept message is a tool response, walk backwards to include
+  // the assistant message that produced the tool_call it belongs to.
+  // Clamp to ensure we never exceed MAX_MESSAGES total.
+  if (sliceStart > 0 && sliceStart < conversationMsgs.length) {
+    const originalStart = sliceStart;
+    while (sliceStart > 0 && conversationMsgs[sliceStart].role === "tool") {
+      sliceStart--;
+    }
+    // If walking back would exceed MAX_MESSAGES, revert and drop the
+    // orphaned tool messages instead (move forward past them).
+    if (systemMsgs.length + (conversationMsgs.length - sliceStart) > MAX_MESSAGES) {
+      sliceStart = originalStart;
+      while (sliceStart < conversationMsgs.length && conversationMsgs[sliceStart].role === "tool") {
+        sliceStart++;
+      }
+    }
+  }
+
+  const truncatedConversation = conversationMsgs.slice(sliceStart);
 
   const result = [...systemMsgs, ...truncatedConversation];
 
@@ -2697,6 +2718,14 @@ export async function startProxy(options: ProxyOptions): Promise<ProxyHandle> {
       // Accepts image as: data URI, local file path, ~/path, or HTTP(S) URL
       if (req.url === "/v1/images/image2image" && req.method === "POST") {
         const img2imgStartTime = Date.now();
+
+        // Cancel the upstream call (and its payment) if the client disconnects.
+        // Registered before preprocessing so source-image fetches also respect it.
+        const clientAbort = new AbortController();
+        res.on("close", () => {
+          if (!res.writableEnded) clientAbort.abort();
+        });
+
         const chunks: Buffer[] = [];
         for await (const chunk of req) {
           chunks.push(Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk));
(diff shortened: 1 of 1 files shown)


VickyXAI (changes requested):
Thanks — both fixes address real gaps (#251, #252), and the branch is green locally (tsc, eslint, prettier, 135 proxy tests). Three things before merge:

**1. The walk-back branch in `truncateMessages()` is dead code.**
Before the walk-back, `systemMsgs.length + maxConversation === MAX_MESSAGES` exactly. Stepping `sliceStart` back by any `k ≥ 1` makes the clamp `systemMsgs.length + (conversationMsgs.length - sliceStart) > MAX_MESSAGES` true for every `k`, so the `while (… sliceStart--)` loop is always reverted and the forward-skip is what actually runs. The *behavior* is right (orphaned `tool` messages get dropped, no 400 from the provider) but the comment promises "include the assistant message that produced the tool_call" and that never happens. Please simplify to the forward skip and describe that — `MAX_MESSAGES` is a BlockRun API limit, so walking back was never an option anyway.

**2. No test.** `truncateMessages` currently has zero coverage. Please add a small `src/proxy.truncate.test.ts` (export the function, or test through the existing `startProxy` harness) with: (a) 250 messages where the boundary lands on a `tool` message → result has no leading `role: "tool"` and `length ≤ MAX_MESSAGES`; (b) boundary lands on an assistant `tool_calls` message → unchanged from today; (c) ≤ `MAX_MESSAGES` → identity.

**3. The img2img outer `catch (err)` (the one that writes `"Image editing failed"` 502) also needs `if (clientAbort.signal.aborted) return;`.** With `signal` now passed to `payFetch`, a mid-flight disconnect surfaces there as an `AbortError`, logs a bogus "Image editing error" and tries to write a 502 to a closed socket. The generations handler you mirrored has exactly this guard on its outer catch — copy it over so the two paths stay identical.

One note for the description, not a blocker: abort prevents settlement only when the disconnect lands before the paid retry is sent; once the 402-retry is in flight the server may still settle. Same caveat as the generations handler.

VickyXAI:
Thanks again for this — both fixes landed on main in `83d1d34` (v0.12.252, on npm now) with the three review changes applied (forward skip only, regression tests in `proxy.truncate-tool-pairs.test.ts` / `proxy.img2img-abort.test.ts`, abort guard on the outer catch). You're credited in the CHANGELOG for #251 and #252. Apologies for the overlap: I landed it directly instead of waiting on your revision. Closing as superseded.

Sertug17:
No worries. I was just pushing the revision when I saw it landed. Glad both fixes are in, thanks for the credit and the thorough review. @VickyXAI

PR #254, opened 2026-08-20
fix(image): stop advertising dall-e-3 and flux-1.1-pro in the OpenClaw image picker

memosr:
## Problem

`buildImageGenerationProvider()` still advertises two image models that the gateway cannot serve:

| id | status |

|---|---|

| `openai/dall-e-3` | gateway returns 400 — delisted upstream 2026-05-25 |

| `black-forest/flux-1.1-pro` | no gateway entry at all |

Both were removed from `IMAGE_PRICING` and `MODEL_ALIASES` in v0.12.227, but the provider list in `src/index.ts` was missed by that sweep.

This is not cosmetic. OpenClaw's image UI sends the picked entry from `provider.models` directly as the `model` field, and the `/v1/images/generations` handler forwards the request body to the gateway **verbatim** — there is no `resolveModelAlias()` pass on that path (unlike `/cr-imagegen`, which does map the `dall-e-3` shorthand to `openai/gpt-image-2`). So picking either entry in the picker is a guaranteed upstream failure.

The inverse gap exists too: the three image models added since that sweep — `openai/gpt-image-2`, `google/nano-banana-2`, `bytedance/seedream-5-pro` — are priced and aliased, but were never advertised, so they are unreachable from the picker.

The `blockrun_image_generation` partner tool description carried the same stale pair, pointing agents at both dead ids and using `openai/dall-e-3` as its `model` param example.

This is the same drift class as v0.12.164 (`buildImageGenerationProvider` advertising only 4 of 8 image models) and v0.12.206 (`azure/sora-2` missing from `VIDEO_PRICING`).

## Changes

- `src/index.ts` — picker list resynced to the 9 ids the gateway serves; `buildImageGenerationProvider` exported so it can be asserted on.

- `src/proxy.ts` — `IMAGE_MODEL_IDS` exported (derived from `IMAGE_PRICING`) as the pinning source of truth; two stale `--model dall-e-3` hints in user-facing copy and a code comment updated to `gpt-image-2`.

- `src/partners/registry.ts` — tool description model list, `shortDescription`, and the `model` param example refreshed.

- `src/index.image-provider.test.ts` — new regression test pinning the picker against `IMAGE_PRICING`, plus explicit assertions on the delisted ids and the live successors.

The `IMAGE_MODEL_ALIASES` entry mapping the `dall-e-3` shorthand to `openai/gpt-image-2` on the `/cr-imagegen` path is deliberate backward compatibility and is left untouched.

## Note on the implementation

The picker list is kept as an explicit array rather than spread from `IMAGE_MODEL_IDS`. Importing that symbol into `src/index.ts` breaks `src/index.lifecycle.test.ts`, which fully mocks `./proxy.js` and would need the new export added to its mock. The test pins the two lists together instead, which keeps the guard without touching an unrelated test's mock surface.

Happy to switch to the derived form and update the lifecycle mock if you prefer that shape.

## Verification

The new test fails on the pre-fix list and passes after:

× only advertises models the gateway can still serve
expected [ 'openai/dall-e-3', 'black-forest/flux-1.1-pro' ] to deeply equal []
× does not advertise models delisted upstream
× advertises the live successors

Full suite after the fix: **721 passed | 3 skipped**. `tsc --noEmit`, `eslint src/`, and `prettier --check src/` all clean.

Note: the 400 on `openai/dall-e-3` is taken from the v0.12.227 changelog entry and the removal of both ids from `IMAGE_PRICING`/`MODEL_ALIASES` in that release; I did not re-probe the live gateway.

## Summary by CodeRabbit

* **New Features**
  * Added support for Nano Banana 2, GPT Image 2, and Seedream 5 Pro image models.
  * Exposed the image generation provider for programmatic use.
  * Updated image-generation commands and examples to use GPT Image 2.

* **Bug Fixes**
  * Removed retired DALL-E 3 and Flux 1.1 Pro models from supported model listings.
  * Improved guidance shown when image uploads fail.

* **Tests**
  * Added validation to ensure advertised image models are current, supported, and include the default model.


memosr:
Both applied in 6c99b1a, thanks.

The `nano-banana-2` alias gap was real — added `banana-2` and `nano-banana-2` to `IMAGE_MODEL_ALIASES` plus a row in the `/cr-imagegen` help block. The shorthand was previously passed through raw and 400'd while the full id worked.

On the catalog drift: the test now asserts set equality with `IMAGE_MODEL_IDS` in both directions, so a model that is priced but not advertised fails too. I left the picker list as an explicit array rather than spreading `IMAGE_MODEL_IDS` — importing that symbol into `src/index.ts` breaks `src/index.lifecycle.test.ts`, which fully mocks `./proxy.js` and would need the new export added to its mock. Happy to switch to the derived form and update the lifecycle mock if a maintainer prefers that shape.

Full suite: 721 passed | 3 skipped. typecheck, eslint, prettier all clean.

VickyXAI:
Merged and shipped in [v0.12.247](https://github.com/BlockRunAI/ClawRouter/releases/tag/v0.12.247), live on npm — thank you @memosr! 🙏

This was a model contribution: you caught drift in *both* directions (two dead ids advertised, three live models unreachable), diagnosed the exact failure path (verbatim forward, no alias resolution), pinned it with a regression test so it can't rot again, and turned the review feedback around fast.

Your PR also kicked off a deeper sweep on our side — we live-probed the gateway and your premise held perfectly (dall-e-3/flux reject pre-payment, all 9 advertised models quote). The probes surfaced a few adjacent gaps we fixed in the same release: orphaned dall-e-3 sizes still in the size picker, six missing Seedream sizes, and `/cr-imagegen` not polling slow-model 202 jobs. You're credited in the CHANGELOG. Hope to see you around here again 🦞

PR #273, opened 2026-08-29
fix(dedup): strip timestamps from array-form content blocks

ygd58:
## What

`stripTimestamps` in `src/dedup.ts` and `normalizeForCache` in `src/response-cache.ts`
only strip OpenClaw's injected timestamp prefix when a message's `content` is a plain
string. Anthropic-style content blocks (`content: [{type: "text", text}, {type: "image_url", ...}]`),
used for vision/multimodal messages, were never handled — the timestamp lives in the
leading text block's `text` field, not in `content` itself, so neither function's
string-only branch ever fires for these.

## Impact

OpenClaw injects a fresh `[DAY YYYY-MM-DD HH:MM TZ]` timestamp on every request. For a
retried multimodal message (timeout, network blip, etc.):

- **Dedup** (`dedup.ts`): the retry hashes differently from the original, so it's never
  recognized as a duplicate — the same request can be paid for twice via x402.
- **Response cache** (`response-cache.ts`): the cache key never matches on retry, so it
  always misses and re-hits the upstream LLM unnecessarily.

Reproduced in isolation before fixing — same-content requests differing only by
timestamp hashed identically for plain-string `content`, but differently for array-form
`content`.

## Fix

Both functions now also walk array-form `content`, stripping the timestamp prefix from
any `{type: "text", text}` block, mirroring the existing plain-string behavior.

## Testing

- Added `src/dedup.test.ts` (new file) — 3 tests covering string content (existing
  behavior, unchanged), array content (the fix), and confirming array content that
  actually differs still produces different keys.
- Added 1 test to `src/response-cache.test.ts` covering the same array-content case.
- Full suite passes: `npx vitest run` → all tests green, no regressions.
- `tsc --noEmit`, `eslint`, and `prettier --check` all clean on the changed files.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved request deduplication for multimodal messages with array-based content.
  * Improved response caching by recognizing equivalent requests with timestamped leading text blocks.
  * Preserved meaningful timestamps in later content blocks to avoid incorrect request matching.
  * Reduced unnecessary duplicate requests and repeated cache misses for multimodal content.

* **Tests**
  * Added coverage for timestamp normalization in deduplication and response-cache scenarios.


VickyXAI:
Thanks @ygd58 — this is a real bug, and the PR nailed the diagnosis: the string-only branch never fires for array-form content, so multimodal retries could be paid twice. Verified locally: full suite green, and the strip functions feed key computation only (the proxy forwards the original body untouched).

Review (including two independent adversarial passes) converged on one scope issue, so I pushed a follow-up commit to your branch (ab325ba) rather than round-tripping:

**Stripping was applied to *every* text block, but OpenClaw only injects into the first one.** A bracketed timestamp starting a *later* text block is user data (e.g. a pasted log line like `[Wed 2026-08-27 09:15 UTC] connection refused`). Stripping it let two genuinely different requests collide on one key — the wrong cached response served for up to 10 min, or a distinct paid request wrongly deduped. Note the plain-string path never had this problem: the injected stamp is always the leading one there, so a user's own timestamp survives the single anchored replace.

The commit scopes the strip to the first `type:"text"` block, extracts the shared `stripLeadingTextBlockTimestamp` helper + `TIMESTAMP_PATTERN` into `src/timestamp-strip.ts` (so dedup and cache normalization can't drift — also addresses the CodeRabbit note), and adds regression tests for the non-leading-block and differing-content cases in both suites. One deliberate residual: `stripTimestamps` still recurses deeper than `normalizeForCache` (pre-existing design difference), but both now share the block-level logic.

Your tests and behavior for the intended case are unchanged and passing. Will credit you in the CHANGELOG on the next release.

PR #278, opened 2026-08-29
fix(proxy): guard image2image abort errors after client disconnect

### LINKED ISSUE #277: Guard image2image abort errors after client disconnect
## Summary

Add client-abort guards to the `/v1/images/image2image` error paths in `src/proxy.ts`.

## Rationale

The handler now passes `clientAbort.signal` to source and mask URL downloads and to the upstream `payFetch` request. When the client disconnects, either operation can reject with an abort error.

The parse catch currently treats an aborted URL download as invalid input and can try to write a `400` response. The outer catch can log the abort and try to write a `502` response. The closed response must not receive either error response.

## Required changes

- In the parse catch that resolves `image` and `mask` input URLs, return immediately when `clientAbort.signal.aborted`.
- In the outer catch around the upstream image2image request, return immediately when `clientAbort.signal.aborted`.
- Preserve the current `400` and `502` behavior for non-abort errors.
- Add Vitest coverage for client disconnect during:
  - source or mask URL download
  - upstream `payFetch`

## Acceptance criteria

- A client disconnect during either path does not log an image-editing error.
- A client disconnect during either path does not call `res.writeHead()` or `res.end()` to send an error response.
- Non-abort failures retain their current error responses.

## Backlinks

- Follow-up from PR #276: https://github.com/BlockRunAI/ClawRouter/pull/276
- Review comment: https://github.com/BlockRunAI/ClawRouter/pull/276#discussion_r3887529499
- Requested by: @Sertug17

Sertug17:
Closes #277.

Follow-up from #276 the `/v1/images/image2image` handler now returns silently when `clientAbort.signal.aborted` is true in both catch paths:

1. **Parse catch** (source/mask URL download): previously wrote a `400` to the closed socket when an abort interrupted the image/mask fetch.
2. **Outer catch** (`payFetch`): previously logged a bogus `Image editing error` and attempted a `502` response after client disconnect.

Non-abort errors retain their current `400` / `502` behavior only abort-caused exceptions are suppressed.

### Checklist
- [x] Relevant issues linked (#277)

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented unnecessary error responses when image-to-image requests are canceled by the client.
  - Improved handling of interrupted requests during request parsing and upstream processing.

### DIFF
--- a/src/proxy.ts
+++ b/src/proxy.ts
@@ -2739,6 +2739,7 @@ export async function startProxy(options: ProxyOptions): Promise<ProxyHandle> {
           img2imgCost = estimateImageCost(img2imgModel, parsed.size, parsed.n || 1);
           reqBody = JSON.stringify(parsed);
         } catch (parseErr) {
+          if (clientAbort.signal.aborted) return;
           const msg = parseErr instanceof Error ? parseErr.message : String(parseErr);
           res.writeHead(400, { "Content-Type": "application/json" });
           res.end(JSON.stringify({ error: "Invalid request", details: msg }));
@@ -2818,6 +2819,7 @@ export async function startProxy(options: ProxyOptions): Promise<ProxyHandle> {
           res.writeHead(200, { "Content-Type": "application/json" });
           res.end(JSON.stringify(result));
         } catch (err) {
+          if (clientAbort.signal.aborted) return;
           const msg = err instanceof Error ? err.message : String(err);
           console.error(`[ClawRouter] Image editing error: ${msg}`);
           if (!res.headersSent) {


VickyXAI:
Landed on main as `50d8447` (v0.12.255) with your authorship preserved. Your branch had forked before v0.12.252, so `clientAbort` wasn't declared in the img2img handler on your base (`tsc`: `Cannot find name 'clientAbort'` — CodeRabbit flagged the same), and the outer-catch guard had already shipped in v0.12.253. Rather than round-tripping a rebase, I applied the remaining parse-catch guard on main and added the #277 regression case (client disconnect during the source download: abort observed, no bogus 400, zero upstream hits). Thanks for the fix — credited in the v0.12.255 CHANGELOG. Closing as landed.

PR #279, opened 2026-08-30
fix(proxy): abort paid upstream on client disconnect for chat path

erhnysr:
## Describe your changes

The media handlers all abort their upstream on client disconnect (#251, fixed for `/v1/images/image2image` in v0.12.252 and for audio + chat-path `/img2img` in v0.12.253). The main `/v1/chat/completions` path — the highest-traffic one — was the one handler still missing it.

`proxyRequest` wired its disconnect abort to `req.on("close")`:

```ts
const onClientClose = () => {
  if (!globalController.signal.aborted) { ... globalController.abort(); }
};
req.on("close", onClientClose);
```

On Node (verified on **v24.15.0**) an `IncomingMessage` emits `"close"` when the **request-body readable finishes**, not when the client hangs up. The body is fully drained near the top of `proxyRequest`, so that event had already fired by the time the listener was attached ~1,500 lines below — `onClientClose` never ran. `globalController` was therefore only ever aborted by the request timeout, and a caller that hung up mid-request left the paid x402 upstream running to completion for a response nobody would receive — the exact orphan-upstream problem #251 fixed for the media paths.

### Change

Move the abort to `res.on("close")` (guarded by `!res.writableEnded`), mirroring every media handler — e.g. `/v1/images/image2image`:

```ts
res.on("close", () => {
  if (!res.writableEnded) clientAbort.abort();
});
```

The four `removeListener` cleanup sites are migrated from `req` to `res` to match.

### Reproduction

New regression test `src/proxy.chat-abort.test.ts` (same harness/style as `src/proxy.img2img-abort.test.ts`): the client POSTs `/v1/chat/completions`, the upstream mock holds the socket open, and the client `destroy()`s mid-request. The mock's `res.on("close")`-with-`!writableEnded` counter observes the abort.

- **Before the fix:** `upstreamAborted` stays `0` — the upstream is orphaned (surfaces as `Response finished with error: Premature close` → `All 1 models failed (HTTP 500)`).
- **After the fix:** the upstream socket observes the abort (`upstreamAborted === 1`), and the test also completes far faster since the abort fires immediately instead of waiting out the premature-close path.

Full suite: 64 files / 724 tests passing.

### Checklist

- [x] Relevant issues linked (#251 lineage; sibling PRs #276 / #278)

## Summary by CodeRabbit

- **Bug Fixes**
  - Client disconnections during chat completions now properly cancel ongoing paid upstream requests.
  - Prevents unnecessary processing and costs after a client connection is closed.

- **Tests**
  - Added regression coverage to verify upstream requests are aborted exactly once when clients disconnect.

- **Chores**
  - Released version 0.12.254.


VickyXAI:
Merged as-is via fast-forward (`fe8e871`) and shipped in v0.12.254 with two small follow-ups on top (`bd13340`): a `res.destroyed` check right after the listener attach (covers a hang-up during the context-compression / balance-check awaits, where `close` had already fired), and a `ClientDisconnectedError` so a client cancel no longer surfaces as `Request timed out after 300000ms` + a balance-cache invalidation. Your test now also asserts `onError` stays silent and gates the disconnect on the upstream being reached instead of a fixed timer. Great diagnosis — thank you!

PR #284, opened 2026-08-31
chore(brand): refresh the snapshot — the markers were rendering a stale catalog

VickyXAI:
`brand-numbers.json` in this repo was behind the published artifact, so every `br:` marker here rendered a stale number.

The markers worked correctly — they rendered the input they had. `--check` is offline by design so PR CI stays deterministic, which means it validates against this repo's **own** snapshot; only `--refresh` updates that snapshot.

Opened automatically by [`brand/fanout-brand-numbers.mjs`](https://github.com/BlockRunAI/blockrun/blob/main/brand/fanout-brand-numbers.mjs). Produced by `--refresh`, no hand-edited digits.

## Summary by CodeRabbit

* **Chores**
  * Updated the recorded `clawrouter.aliases` count from 259 to 229.

### DIFF
--- a/brand-numbers.json
+++ b/brand-numbers.json
@@ -18,7 +18,7 @@
     "dimensions": 15,
     "tiers": 4,
     "profiles": 4,
-    "aliases": 259
+    "aliases": 229
   },
   "mcp": {
     "tools": 20


VickyXAI:
Closing — this snapshot is born stale and would regress the markers.

The fan-out ran during the window between blockrun's alias-count merge and its Cloud Run deploy, so it read the *old* published `/brand/numbers.json` and baked in 229.

Both sources of truth now say **259**:
- `Object.keys(MODEL_ALIASES).length` in `src/models.ts` → 259
- the deployed `https://blockrun.ai/brand/numbers.json` → 259

Merging this would rewrite every `br:clawrouter.aliases` marker in the repo from the correct 259 down to a stale 229. CI already caught it: `AssertionError: expected [...] to have a length of 229 but got 259` in `brand-numbers.test.ts` — the red Build & Test is the test doing its job, not a flake.

The next fan-out will regenerate correctly now that the deploy has landed.

PR #291, opened 2026-08-31
feat(desktop): add unified ClawRouter control plane

KillerQueen-Z:
## Summary

- add the Electron/React ClawRouter Desktop control plane under `apps/desktop`
- support one-click connect/restore flows for OpenClaw, Codex, Hermes, DeepSeek Harness, and Pi
- share the BlockRun Core Base and Solana wallets across the CLI, proxy, and Desktop
- expose detailed local model metadata for prices, context windows, capabilities, and provider branding
- add a wallet-bound Coinbase Onramp flow and Base/Solana balance switching
- stage the current repository build into packaged Desktop releases so the UI and core cannot drift
- document the macOS developer preview and add a dedicated Desktop CI job

## Validation

- ClawRouter tests: 831 passed, 1 skipped
- Desktop tests: 25 passed, 1 skipped
- ClawRouter typecheck and lint: passed
- Desktop typecheck: passed
- Prettier and `git diff --check`: passed
- ClawRouter build and dist smoke test: passed
- Desktop renderer and Electron main-process builds: passed
- full macOS arm64 `.app` packaging: passed
- packaged runtime verified to contain this branch's wallet, model metadata, and Onramp implementation
- npm pack dry run: passed; `apps/desktop` remains outside the existing npm package

## Release note

The generated macOS app is currently an unsigned developer preview. Developer ID signing and notarization remain release-pipeline work; they are not required to review the source integration in this PR.

## Summary by CodeRabbit

- **New Features**
  - Added a macOS desktop preview with dashboards for agents, models, usage, settings, wallet balances, and router health.
  - Added setup and management support for OpenClaw, Codex, Hermes, DSH, and Pi.
  - Added model catalog filtering, capability details, pricing, and context information.
  - Added Base and Solana payment-chain switching and Coinbase funding via validated onramp links.
  - Added shared wallet detection, including BlockRun Core wallet support.
- **Documentation**
  - Added desktop build, packaging, configuration, and supported-agent guidance.
- **Tests**
  - Added coverage for configuration, wallets, onramp validation, security checks, and rollback behavior.


VickyXAI (changes requested):
Thanks for this — it's a lot of careful work, and the Electron side is genuinely well built. I went looking hard for a renderer→main escape and didn't find one: `contextIsolation` on, `nodeIntegration` off, `sandbox: true`, `will-navigate` and `setWindowOpenHandler` both guarded, a CSP meta tag, every IPC handler validated through a parser, config writes at `0o600`, no `dangerouslySetInnerHTML`, and no path where a private key or mnemonic reaches the renderer. The TOML-injection case even has a unit test. That part I'd merge as-is.

What I can't merge yet is the wallet resolution rewrite and the bundled model catalog. Six blocking items below. All of them pass the current suite — one is actively pinned by a new test — so please don't read green CI as disagreement.

---

## Blocking

### 1. The wallet precedence flip silently changes which wallet pays

`src/auth.ts:290-334`

Precedence moves from `saved → env → generate` to `env → core → saved`. Any user whose `~/.openclaw/blockrun/wallet.key` differs from `~/.blockrun/.session` will, on the next start, begin paying from a different wallet. Their funded ClawRouter balance strands and payments fail with insufficient funds while the UI shows a healthy balance for the wrong address.

Reproduced:

```
~/.openclaw/blockrun/wallet.key  = 0x1111...11
~/.blockrun/.session             = 0x2222...22
resolveOrGenerateWalletKey() → source="core", key=0x2222...22
```

These two files are written by independent generators (ClawRouter auto-generates one, `@blockrun/core` the other), so "they're usually the same" isn't safe to assume. The function's own comment a few lines up warns against exactly this class of silent wallet swap:

> File exists but content is wrong — do NOT silently fall through to generate a new wallet. This would silently replace a funded wallet with an empty one.

**Ask:** keep `saved` ahead of `core`. Adopt Core only when there is no legacy wallet, or gate the switch behind an explicit opt-in (`clawrouter wallet use-core`) that prints both addresses and both balances first. Existing installs must not change payer without the user saying so.

### 2. Solana payer split-brain — users are told to fund an address the proxy never spends from

`src/auth.ts:118-138` vs `src/index.ts:1595-1607`, `src/index.ts:1660-1670`, `src/doctor.ts:419-425`

`resolveSolanaWallet()` now prefers `~/.blockrun/.solana-session`, and the proxy signs from that (`src/proxy.ts:2369-2374`) and monitors that balance (`:2427`). But three consumers still derive from the mnemonic:

- `/wallet export` prints the mnemonic-derived Solana address as "your" address
- `/wallet solana` prints `Fund with USDC on Solana: https://solscan.io/account/<mnemonic-derived>`
- `doctor.ts` displays the Core-derived address at line 137, then registers the **mnemonic** signer at line 425 — so it reports one address and pays from another

Reproduced with a mnemonic and a `.solana-session` both present:

```
proxy pays from   FAe4sisG95oZ42w7buUn5qEE4TAnfTTFPiguZUHmhiF
/wallet export shows  3Cy3YNTFywCmxoxt8n7UH6hg6dLo5uACowX3CFceaSnx
```

Solana is the default payment chain for new installs, so this is the default path, not an edge case. Users will fund an address that never gets spent.

**Ask:** every Solana consumer goes through `resolveSolanaWallet()`. No second derivation site.

### 3. A malformed Core Solana session bricks startup for Base-only and free-tier users

`src/auth.ts:291` calls `resolveSolanaWallet()` unconditionally, and it throws on a non-ENOENT read error, a malformed `.solana-session`, or a malformed `SOLANA_WALLET_KEY`.

Reproduced with `payment-chain=base` and a corrupt session file:

```
Cannot load the BlockRun Core Solana wallet at ~/.blockrun/.solana-session:
Solana key contains a non-base58 character.. Refusing to fall back to a different wallet.
```

That user never asked for Solana. Free-tier users need no wallet at all and still can't start. Before this PR, `loadMnemonic()` explicitly "warns on corruption but never throws" for this reason.

`src/auth.wallet-core.test.ts:60-64` asserts the throw is correct, so the test pins the behavior rather than catching it.

**Ask:** resolve Solana lazily, or only when the payment chain is `solana`. Refusing to silently swap wallets is right; refusing to start is not the same thing. Degrade to "no Solana key, here's why" and let Base keep working.

### 4. `/wallet export` hands out a mnemonic that can't recover the wallet in use

`src/auth.ts:337-344`

In the generate branch, `mnemonic: result.mnemonic` (freshly created) is returned alongside `solanaPrivateKeyBytes: solana?.privateKeyBytes ?? result.solanaPrivateKeyBytes` (Core's, when present). The two disagree. `/wallet export` then prints:

> CRITICAL: Back up this mnemonic. It is the ONLY way to recover your Solana wallet.

Which is false in that state. That's a permanent-fund-loss instruction.

**Ask:** only return a mnemonic when it actually derives the Solana key that's in use, and have the export path say where the key came from.

### 5. `model-catalog.ts` is a ~40-release-old snapshot, and it wins on the paths that matter

`apps/desktop/electron/core/model-catalog.ts` (2,441 lines)

The snapshot predates v0.12.221 — it still names `fable → Claude Opus 4.8`. Diffed against `BLOCKRUN_MODELS` + `buildDetailedModelList()`: **197 field disagreements across 110 ids.**

Prices, all money-facing:

| id | bundled in/out | registry in/out |
|---|---|---|
| `deepseek/deepseek-v4-pro` | 0.435 / 0.87 | **1.32 / 3.96** |
| `google/gemini-3.5-flash` | 0.5 / 3 | **1.5 / 9** |
| `fable`, `fable-5`, `fable-5.0` | 5 / 25 | **10 / 50** |
| `gpt5` | 5 / 30 | **2 / 12** |
| `grok` | 1.5 / 4 | **2.5 / 9** |
| `zai/glm-5` | 0.6 / 1.92 | **1 / 3.2** |

Also: 28 registry models missing entirely, including `anthropic/claude-opus-5`, `anthropic/claude-fable-5`, `anthropic/claude-sonnet-5`, the whole `openai/gpt-5.6-*` family, and 6 of 7 free-tier models including `free/nemotron-3.5-lightning` — the current `free` alias target. It binds bare `qwen` to flagship pricing at `model-catalog.ts:1577`, a name the registry deliberately leaves unbound so callers expecting a free `qwen*` model don't get billed. And it reintroduces the `gemini-2.5-flash-lite` `vision: false` bug that was fixed in `fea53e8`, one commit before this branch's base.

This isn't a rare fallback path:

- Every field is `live ?? bundled` (`manager.ts:232-241`, `App.tsx:1228-1241`), and `/v1/models` carries no metadata at all (`proxy.ts:1580-1592`). So against any proxy that predates this PR — i.e. every proxy shipped to date, exactly the "older running ClawRouter proxies" the file header names — all ten fields come from the snapshot.
- `adapters/pi.ts:80` fetches `${proxyBaseUrl}/models`, never `/admin/models`, and `toPiModel` (`pi.ts:95-115`) resolves context window, max output, cost and vision 100% from `BUNDLED_MODEL_METADATA` — then writes them into the user's real `~/.pi/agent/models.json`. Stale prices get persisted into an agent config.

There's no generator, no test, and no CI check. The header comment "Regenerate whenever the root model catalog changes" is the entire enforcement mechanism, and this repo has lost that bet before.

**Ask:** delete it and hard-require `/admin/models`. `apps/desktop/runtime/` already stages the repo's own build, so the desktop app controls which proxy it talks to and the "older proxy" case doesn't arise. If a bundle is genuinely needed, it needs `scripts/gen-model-catalog.mjs` plus a vitest that fails when it diverges from `buildDetailedModelList()`. Either way `pi.ts` must read `/admin/models`.

### 6. The packaged app ships two ClawRouter cores; only one gets staged

`apps/desktop/runtime/pnpm-lock.yaml:155` and `:165`

The lock pins both `@blockrun/clawrouter@0.12.212` (transitive via `@blockrun/clawrouter-codex@0.4.0`) and `@blockrun/clawrouter@0.12.261` (direct), with distinct integrity hashes — two separate on-disk copies under pnpm's isolated layout.

`scripts/stage-root-runtime.mjs:7,18` stages exactly one path:

```js
const packageLink = join(desktopRoot, "runtime", "node_modules", "@blockrun", "clawrouter");
```

So the Codex path loads an unstaged v0.12.212 — about 50 releases old, predating the `max_tokens` / `max_completion_tokens` cost-cap fix in v0.12.232. The README's claim that this "keeps the Desktop release and the repository commit on exactly the same wallet, routing, and model implementation" doesn't hold for Codex.

**Ask:** stage every nested copy, or bump `clawrouter-codex` to a release whose peer range accepts 0.12.261 so pnpm dedupes to one.

---

## Non-blocking, but worth fixing before release

- `electron/core/config.ts:26-30` — the Codex root block is blind-prepended after stripping only ClawRouter's own markers. A user who already sets root `model` / `model_provider` (the documented way to set a Codex default) ends up with duplicate root keys and unparseable TOML, so Codex won't load its config at all. `codex.ts:50-51` only regexes for the provider table, so `verify()` passes and the transaction never rolls back. Merge into the parsed document and make `verify()` actually parse.
- `electron/core/runtime.ts:68-77` and `adapters/openclaw.ts:50` — a button press runs `npx -y ...@latest` / `npm install ...@latest` with lifecycle scripts enabled, bypassing the pinned runtime the app already ships as `extraResources`. Only `pi.ts:43-45` passes `ignoreScripts`. Prefer `bundledBin()`, pin versions, and pass `--ignore-scripts` everywhere.
- `electron/core/manager.ts:96-102` — reports `rolledBack: true` on every failure, including when nothing was snapshotted and when `restoreFiles` itself threw. It also only covers `adapter.managedPaths`, so `clawrouter setup`, the pip install, and the npm installs are never reverted.
- `electron/core/manager.ts:213-216` with `App.tsx:54` — wallet addresses are sent to `mainnet.base.org` and `api.mainnet-beta.solana.com` every 15s for the life of the process (`main.ts:50-55` hides rather than closes the window), while Settings advertises "Secrets stay local" (`App.tsx:627`). Read balances from the local proxy, or gate on window visibility and disclose the endpoints.
- `src/api.ts:96-108, 212-214` — if the preload bridge fails to load, `demoDashboard` renders `balance: 18.42` and fake addresses as real, and `fallbackApi.install` reports success for operations that never happened. Render an explicit error state instead.
- `electron/core/transaction.ts:28-32` — `~/.openclaw/agents/main/agent/auth-profiles.json` is a credential store that can hold real third-party API keys; `snapshotFiles` base64s it into a second copy at `~/.clawrouter-desktop/backups/` that persists until `restoreOriginal`. Mode is `0o600` (good), but please exclude credential stores or store a hash and mode only.
- `src/App.tsx:527` — `const values = [42, 68, 53, 82, 64, 91, 74]` renders under "Routing activity — Requests handled by the local proxy". Fabricated data presented as telemetry. CodeRabbit flagged this too.
- `src/cli.ts:913-923` — `clawrouter onramp` funds Base with no payment-chain check, but new installs default to Solana. Warn (or refuse) when the active chain is Solana; the gateway is Base-only server-side so the CLI has to be the one to say so.
- `src/wallet.ts:93-140` — hand-rolled base58 decoder on a private-key path. `@scure/base` is already in the dependency tree and exports `base58`. Please use it.
- `.github/workflows/ci.yml:74` — the desktop job pins Node 24 while every other job and `engines` say 22, and it never runs `npm run dist`. The staging step that finding 6 is about is completely untested in CI.
- `README.md:82` — Option A puts an unsigned, build-from-source Electron app above the working `npm install -g` path, without mentioning that macOS will refuse to open the result without a Gatekeeper bypass. Until there's a signed download, I'd keep OpenClaw as Option A and add Desktop below it.
- `adapters/hermes.ts:52-60` — falls back to `python3 -m pip install` against whatever `python3` PATH resolves to, unpinned, no venv, no `--user`, and outside the transaction snapshot so uninstall leaves it behind.

---

## Suggested path

The desktop app is worth landing. The wallet rewrite is what's holding it up, and it's separable — findings 1-4 are contained in `src/auth.ts` plus three call sites, and 5-6 are about deleting a snapshot rather than adding code.

If it helps, splitting this into two PRs would move faster: one for the shared-wallet work (small, needs careful review, changes behavior for existing users), and one for `apps/desktop` (large, additive, low risk once it stops shipping its own model prices). Happy to review either as soon as it's up.

VickyXAI:
Superseded by #313, which carries the Desktop control plane forward on a branch that merges cleanly and drops the wallet-precedence rewrite this PR carried. Closing so the review history stays in one place — the work and the credit continue there.

PR #294, opened 2026-09-01
fix(polymarket): reject non-positive amount_usd in trade/withdraw

erhnysr:
Fixes an input-validation gap in `executeTrade()` (market buy path) and `withdrawFunds()`: neither rejected a non-positive `amount_usd`, unlike the sibling `fund.ts`.

- `orders.ts`: a negative `amount_usd` becomes a negative order notional. `reserveBet()` does `ledger.totalUsd += notional`, so a negative notional *decreases* the running total instead of increasing it — letting `ledger.totalUsd + notional > sessionCap` stay satisfied indefinitely and silently defeating `POLYMARKET_MAX_SESSION_USD` for every subsequent order.
- `withdraw.ts`: `amountRaw > balanceRaw` is false for zero/negative amounts, so a non-positive `amount_usd` passes the balance check and reaches the bridge transfer call.

Both are fixed with the same guard pattern already used in `fund.ts`. Added two new tests to `polymarket.test.ts` (market-buy path) and a new `withdraw.test.ts` (withdraw path); verified failing before the fix, passing after.

## Summary by CodeRabbit

- **Security**
  - Trades and withdrawals are now checked against configured spending policies before authorization.
  - Rejected trades release their reserved session balance.

- **Bug Fixes**
  - Purchases and withdrawals reject invalid, non-finite, zero, or negative dollar amounts before processing.
  - Limit orders reject invalid or non-positive sizes before network requests.
  - Invalid requests no longer trigger network calls.

- **Tests**
  - Added coverage for invalid purchase amounts, limit-order sizes, and withdrawal amounts.

### DIFF
--- a/src/polymarket/orders.ts
+++ b/src/polymarket/orders.ts
@@ -289,6 +289,23 @@ export async function executeTrade(input: TradeInput): Promise<ToolResult> {
       isError: true,
     };
   }
+  // A non-positive amount_usd would otherwise become the order's notional
+  // (see below) and, being additive, a negative value lets the ledger check
+  // (`ledger.totalUsd + notional > sessionCap`) be satisfied by *reducing*
+  // totalUsd instead of raising it — silently defeating POLYMARKET_MAX_SESSION_USD
+  // for every order placed afterward. Reject before any network call, same as
+  // fund.ts's amount_usd guard.
+  if (
+    !isLimit &&
+    input.action === "buy" &&
+    input.amount_usd !== undefined &&
+    input.amount_usd <= 0
+  ) {
+    return {
+      text: `amount_usd must be a positive dollar amount to spend, got ${input.amount_usd}.`,
+      isError: true,
+    };
+  }
   if (!isLimit && input.action === "sell" && input.size === undefined) {
     return { text: `Market sells need size (shares to sell).`, isError: true };
   }
--- a/src/polymarket/polymarket.test.ts
+++ b/src/polymarket/polymarket.test.ts
@@ -112,4 +112,21 @@ describe("trade gating (cross-field validation, no network)", () => {
     expect(r.isError).toBe(true);
     expect(r.text).toMatch(/GTD orders need expires_at/i);
   });
+
+  // A negative or zero amount_usd must be rejected here, before any network
+  // call: it becomes the order's notional, and being additive it lets
+  // `ledger.totalUsd + notional > sessionCap` be satisfied by *reducing*
+  // totalUsd instead of raising it — silently defeating
+  // POLYMARKET_MAX_SESSION_USD for every order placed afterward.
+  it("rejects a market buy with a negative amount_usd before touching the CLOB", async () => {
+    const r = await executeTrade({ action: "buy", token_id: "1", amount_usd: -5 });
+    expect(r.isError).toBe(true);
+    expect(r.text).toMatch(/amount_usd must be a positive dollar amount/i);
+  });
+
+  it("rejects a market buy with amount_usd of zero before touching the CLOB", async () => {
+    const r = await executeTrade({ action: "buy", token_id: "1", amount_usd: 0 });
+    expect(r.isError).toBe(true);
+    expect(r.text).toMatch(/amount_usd must be a positive dollar amount/i);
+  });
 });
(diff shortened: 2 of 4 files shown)


VickyXAI (commented):
Verified every claim in the description against the code rather than taking it — the mechanism is exactly as you describe. `reserveBet()` is `ledger.totalUsd += usd` with no validation (`orders.ts:41`), market-buy notional is `input.amount_usd` verbatim (`orders.ts:359-362`), and both `notional > maxBet` and `totalUsd + notional > sessionCap` are false for a negative, so it walks straight through. `withdraw.ts` likewise: `amountRaw > balanceRaw` is false for non-positive.

Probed with the verbatim expressions:

```
-5       → REJECTED ✓
0        → REJECTED ✓
5        → passes, cap works ✓
Infinity → rejected by the per-order cap ✓
```

Both guards do what they claim. `tsc` clean, 813 tests pass. Two gaps I'd like closed before merge, same class as what you're fixing.

### 1. `NaN` and strings walk through `<= 0`

There is no runtime validation anywhere in `src/polymarket/` — no zod, no `safeParse`. `tool.ts:128` takes `params: Record<string, unknown>` and every field is a bare `as number | undefined`, which is compile-time only. So whatever the model emits in the tool-call JSON arrives raw:

```
NaN     → passes → ledger.totalUsd = NaN     → next $999 order blocked? false
'abc'   → passes → ledger.totalUsd = "0abc"  → next $999 order blocked? false
'5'     → passes → ledger.totalUsd = "05"    (string concat; ledger is now a string)
```

`NaN <= 0` and `"abc" <= 0` are both false, so the guard lets them by; then `totalUsd += NaN` poisons the ledger and every later `totalUsd + notional > sessionCap` is false permanently. That is strictly worse than the negative case you fixed — permanent cap defeat rather than per-order.

`Number.isFinite(x) && x > 0` closes it in both files.

### 2. Limit orders are not covered

The guard is `!isLimit && action === "buy"`. For limit orders notional is `price * size`, and the negative-size check is:

```ts
if (isLimit && minSize > 0 && (size as number) < minSize)
```

with `const minSize = parseFloat(book.min_order_size || "0")` (`orders.ts:329`). A market whose book omits `min_order_size` gives `minSize = 0`, the guard short-circuits, and `size: -100 @ price 0.5` yields `notional = -50` straight into `reserveBet(-50)` — the same ledger-defeating mechanism, different input.

### Note on #304

#304 wraps these signing paths in `signUnderSpendPolicy`, and its canonical-amount check (`/^\d+$/`) would reject a `NaN`/negative notional — but only when the operator has configured an amount limit. Your guard is what protects everyone who hasn't. Both are wanted; they are not redundant.

VickyXAI:

VickyXAI:
Correction to my note above — I had the mechanism wrong.

CI checks out the **merge ref** (your head merged with `main` as of trigger time), not the PR head. So you do *not* need to merge `main` in: any new push retriggers and picks up the fix automatically. What does **not** work is re-running the existing run — that replays the original merge ref and stays red.

#304 is the proof: it pushed a commit at 02:13 (after #300 merged at 02:09) and went green without ever containing `a58c2ae`. Sorry for the misdirection.

VickyXAI (approved):
Good catch, and the mechanism in your description was exactly right — I verified every step of it against the code before agreeing.

Rather than round-trip, I pushed the two gaps myself as `ade76fc`, along with the `main` merge (your `withdraw.ts` guard conflicted with #304, which had just added a `deps` param — kept both).

What I changed:

- `Number.isFinite(x) && x > 0` instead of `<= 0`, in both files. Nothing validates these at runtime: `tool.ts` hands them over as `params.amount_usd as number | undefined`, a compile-time cast over `Record<string, unknown>`, so NaN and strings arrive intact and `NaN <= 0` / `"abc" <= 0` are both false. That case is worse than the one you fixed — a NaN notional poisons `ledger.totalUsd` permanently, since `totalUsd + NaN > cap` stays false forever.
- Added the limit-order path. notional is `price * size` there, and the existing `minSize` check can't catch a negative: `book.min_order_size` is often absent, so `parseFloat(undefined || "0")` is 0 and the `minSize > 0` term short-circuits the whole guard. Put it in the cross-field block before any network call, matching your own fail-closed placement.
- 3 regression tests. I initially placed the size guard next to `minSize` and the test failed with "No Polymarket deposit wallet configured" — it sat after the wallet fetch and was unreachable. Moving it early fixed it, which is a decent argument for that placement on its own.

830 tests pass. Thanks for this one.

PR #299, opened 2026-09-01
fix(models): add paid DeepSeek V4 Flash to picker and sync modelPolicy.allow

0xCheetah1:
## Summary

- Add the paid `deepseek/deepseek-v4-flash` to `top-models.json` and the model registry. The free `deepseek-v4-flash` tier was EOL'd upstream (NVIDIA hosting rot, 410s since 2026-08-12); the paid route survives but is not advertised in `GET /v1/models`, so it fell out of both pickers.
- Sync `agents.defaults.modelPolicy.allow` to `TOP_MODELS`. OpenClaw 2026.8+ gates the `/model` picker on `modelPolicy.allow`, which nothing synced, so newly-shipped models showed in the catalog but were rejected with "not available for this agent". Non-blockrun entries are preserved.

## Validation

- Typecheck: pass
- Build + dist smoke: pass
- Full unit suite: 812 passed (3 new tests for the modelPolicy.allow sync)

## Notes

- Verified live on testbed: the paid `deepseek/deepseek-v4-flash` routes (x402) and is now selectable via `/model`.

## Summary by CodeRabbit

- **New Features**
  - Added DeepSeek V4 Flash to the available BlockRun model catalog, including tool-calling support and expanded context capacity.
  - Model configuration now automatically includes current BlockRun models while preserving custom entries and removing outdated ones.

- **Bug Fixes**
  - Improved handling of missing or invalid model allowlists so configuration updates complete reliably.

- **Tests**
  - Added coverage for model synchronization, stale entry removal, custom entry preservation, and invalid configuration values.


VickyXAI (changes requested):
Two blocking issues. The `modelPolicy` one is proven against a real OpenClaw, not inferred, and it is invisible to CI — `tsc` is clean and 812 tests pass on this branch.

### 1. The `modelPolicy` write bricks every OpenClaw below 2026.8

`src/index.ts:461-466` writes `agents.defaults.modelPolicy` with no OpenClaw version detection (`grep -n -i version src/index.ts` finds only ClawRouter's own `VERSION` import). `agents.defaults` is a strict schema. Copying a real `openclaw.json` to a throwaway `HOME`, injecting the key, and running the installed OpenClaw `2026.5.2 (8b2a6e5)`:

```
Config invalid
File: ~/.openclaw/openclaw.json
Problem:
  - agents.defaults: Unrecognized key: "modelPolicy"

Run: openclaw doctor --fix
```

Every config-loading command dies. `openclaw doctor --fix` does strip the key, but `injectModelsConfig` re-runs on the next gateway start and re-adds it, so the brick loops rather than being a one-time break.

Note `openclaw --version` does **not** parse the config and is useless as a probe here; `openclaw models` is the cheap validator.

This is the same failure class the function's own comments already document a few lines below — the `tools.web.search.provider` validator trap that forced the "set `enabled`, DELETE `provider`" dance.

`openclaw@2026.8.2` does exist on npm (it is `latest`), so the premise is plausible — I could not confirm 2026.8.2 accepts the key because it requires node `>=22.22.3`. So the fix is a version guard, not deletion, and it has to fail **safe**: skip the write when the version is unknown or undetectable. Today it fails open onto a hard config brick.

Related: `src/index.model-policy.test.ts:57,79,89` assert the write happens unconditionally, with no version dimension. They pin the current behaviour, so they need rewriting alongside the fix.

### 2. The fallback discards sibling `modelPolicy` keys

```ts
if (!Array.isArray(modelPolicyAllow)) {
  modelPolicyAllow = [];
  (defaults.modelPolicy as Record<string, unknown>) = { allow: modelPolicyAllow };
```

This replaces the whole object, and it triggers whenever `.allow` is missing *or* non-array. A user with `modelPolicy: { deny: [...] }` and no `allow` silently loses their deny list on the next gateway start. Should merge: `{ ...existing, allow: [] }`.

### 3. `deepseek/deepseek-v4-flash` duplicates a model already in the picker

blockrun is source-of-truth and has no such model id — it is an alias (`blockrun/src/lib/models.ts:2994`):

```js
// deepseek/deepseek-v4-flash existed briefly in commit 4de12e5 before
// we removed the paid V4 Flash SKU to avoid confusion with the free
// nvidia/deepseek-v4-flash.
"deepseek/deepseek-v4-flash": "deepseek/deepseek-chat",
```

ClawRouter already carries that target, byte-identical:

```
deepseek/deepseek-v4-flash   name="DeepSeek V4 Flash"       $0.14/$0.28 ctx=1048576 out=65536   ← new here
deepseek/deepseek-chat       name="DeepSeek V4 Flash Chat"  $0.14/$0.28 ctx=1048576 out=65536   ← already in top-models.json
```

So the picker lists one upstream route twice, which is the confusion blockrun removed the SKU to prevent. The live gateway confirms it: `/v1/models` (98 entries) lists `deepseek/deepseek-chat` and not `deepseek/deepseek-v4-flash`.

"Verified live on testbed: it routes" is true and consistent with this — it routes *because the alias resolves to deepseek-chat*.

### One correction to something I nearly filed against this PR

I initially believed the shorthand `deepseek-v4-flash` pointed at a dead model, because probing the gateway with ClawRouter's local id `free/deepseek-v4-flash` returns `Unknown model`. That was the wrong test: `proxy.ts:toUpstreamModelId()` rewrites `free/X` → `nvidia/X` before anything goes on the wire. Probing the id that is actually sent returns HTTP 200 (served by `nvidia/nemotron-3-super-120b`), so the alias is fine and `models.ts` is right that explicit pins stay routable. Not a defect against you.

What does remain is a naming point worth a thought: after this PR, `/model deepseek-v4-flash` (free tier) and `/model deepseek/deepseek-v4-flash` (the paid entry you are adding) differ only by prefix and resolve to different models.

### Housekeeping

No CHANGELOG entry and no version bump — `package.json` is `0.12.263` on both `main` and this branch, and the only CHANGELOG touch is a prettier reflow of the existing v0.12.263 section.

VickyXAI:
`deepseek/deepseek-v4-flash` is a gateway alias for `deepseek/deepseek-chat`, not a separate SKU — it is not in the model catalog and quotes the same $0.002. Adding it as a registry entry would list the same model twice under two ids.

The second half does not hold either: `agents.defaults.modelPolicy.allow` does not exist in OpenClaw (grepped 2026.9.2 — no matches). The `/model` picker gates on `agents.defaults.models`, which `src/index.ts` already syncs to `TOP_MODELS` on every gateway start, including the stale-entry prune.

PR #303, opened 2026-09-02
feat(policy): `clawrouter policy` CLI + `/policy` plugin command over SpendControl

### LINKED ISSUE #301: SpendControl: a command surface for `setPolicy`/`clearPolicy` — hand-editing `spending.json` is the only operator path
Follow-up to the scope note on #268. On `main` (`fa0b1468`, v0.12.263) spend policy is set only through the library — `SpendControl.setLimit/clearLimit/setPolicy/clearPolicy`, `src/spend-control.ts:337–370`. Nothing in `src/cli.ts` (unchanged since `07a8a384`) or the plugin commands registered in `src/index.ts:1979–2228` calls them. The operator path is hand-editing `spending.json` and restarting, because limits load once in the constructor (`spend-control.ts:331–335`) and are never re-read.

**Proposal** — one handler behind `clawrouter policy …` and `/policy …`, shaped like `/exclude add|remove|clear` (`src/commands/exclude.ts`, `index.ts:2228`):

```
policy                              # print limits + lists on disk
policy set|add|remove <list> <v>…   # <list> = exact PolicyList name
policy clear <list>
policy limit <window> <usd>|clear   # <window> = exact SpendWindow
```

**Fail closed on input; on any refusal exit non-zero and write nothing:**
- `allowedNetworks` values must be CAIP-2. `base`/`solana` refuse, and the message names `CAIP2_BASE`/`CAIP2_SOLANA_MAINNET`. No nickname translation.
- `0x`-prefixed values that are not 40 hex refuse — a typo'd denylist entry is fail-open.
- `<usd>` parsed by `/^\d+(\.\d+)?$/`, never `parseFloat` (`"5abc"` → 5).
- `remove` that empties a list routes to `clearPolicy`: `setPolicy` rejects `[]` (`:356`) while `load` treats it as absent (`:209`).
- Malformed file (`policyFileBroken`, `:327`): `save()` returns silently (`:681`), so a command built on `setPolicy` would report success and write nothing. Expose that state — `getStatus()` (`:641`) does not — and exit non-zero naming the field.

**Live proxy:** re-read limits at `check()` entry when the file mtime changes, or have the command say a restart is required. "Applied" with neither is the worst outcome.

**Tests:** extend `src/proxy.spend-policy.test.ts` — inject `spendControl`, run the handler against the same storage after `startProxy`, assert `signerCalls === 0` for the newly blocked payee; deleting the reload path must fail it. Unit: nickname refused with the file byte-identical; malformed file → non-zero exit, file untouched.

PR to follow.

twzrd-sol:
Closes #301.

## Policy CLI + /policy plugin command over SpendControl

`SpendControl.setLimit/clearLimit/setPolicy/clearPolicy` were library-only:
`src/cli.ts` had no policy subcommand and the plugin registered nothing
policy-shaped, so the only way to configure counterparty lists or spend caps
was to hand-edit `spending.json`. This adds one implementation behind two
entry points, so validation lives in exactly one place.

    policy                                 show limits + lists on disk
    policy set|add|remove <list> <v>...    <list> = exact PolicyList name
    policy clear <list>
    policy limit <window> <usd>|clear      <window> = exact SpendWindow name

Surfaces:

- CLI: `clawrouter policy ...` (src/cli.ts:46-49, src/cli.ts:88, dispatch at
  src/cli.ts:898-906). `parseArgs` treats everything after `policy` as
  policy arguments (src/cli.ts:792-796). Non-zero exit on any rejection.
- Plugin: `/policy` registered via `createPolicyCommand()`
  (src/index.ts:2230), `requireAuth: true`, matching `exclude.ts`.
- Library: `runPolicyCommand(argv, { openControl? })` in
  src/commands/policy.ts:169-214; `openControl` is injectable for tests.

### Input validation (fail-closed, validate before the store is opened)

- All argv parsing and validation happens in `parsePolicyArgs`
  (src/commands/policy.ts:88-127) before `openControl()` is called, so a
  rejection writes nothing.
- `allowedNetworks` entries must be exactly `eip155:8453` or
  `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdpKuc147dw2N9d`
  (src/commands/policy.ts:29-30, 75-80). These are the only networks the
  proxy can pay on, and `SpendControl.check()` matches networks by exact
  string (src/spend-control.ts:426-441), so any other value — including
  valid CAIP-2 ids and nicknames — can never match a quote and would only
  block payments. The underlying store remains permissive; this constraint
  is operator-surface only.
- `0x`-prefixed entries must be `0x` + 40 hex
  (src/commands/policy.ts:33, 81-83).
- Amounts must match `/^\d+(\.\d+)?$/` and be > 0
  (src/commands/policy.ts:32, 104-109); `"5abc"` is rejected, not truncated.
- EVM address entries are lowercased and deduped via `normalizePayee`
  (src/commands/policy.ts:136), matching storage normalization
  (src/spend-control.ts:70-72).
- `remove` that empties a list routes to `clearPolicy()`; `setPolicy([])`
  throws by design (src/commands/policy.ts:144-151).
- `remove` of an absent entry is refused with the missing values named;
  nothing is written (src/commands/policy.ts:145-147).

### Persistence traps closed

Two existing behaviors made a successful setter return compatible with
nothing being written:

1. `policyFileBroken` is private and `save()` silently no-ops while it is
   set (src/spend-control.ts:690-693). New `SpendControl.getPolicyFileError()`
   (src/spend-control.ts:383-385) exposes it; the command refuses with the
   malformed-file reason and exit 1, file untouched
   (src/commands/policy.ts:176-181).
2. `FileSpendControlStorage.save()` logs and swallows write failures
   (src/spend-control.ts:260-262). After every mutation the command reopens
   the store through the real load path and compares the single key it
   changed; a mismatch is reported as "Write did not land" with the on-disk
   value, instead of a false success (src/commands/policy.ts:202-210). The
   single-key scope avoids false positives on unrelated concurrent changes
   (the proxy's `saveHistory()` preserves on-disk limits by design,
   src/spend-control.ts:271-282).

Known limits of the verification: it is read-your-writes on the same host,
not durability (save is write-then-rename without fsync); a concurrent
writer touching the same key between write and read-back produces a
fail-closed "Write did not land", never a false success.

### Applying changes

The proxy reads limits once in the `SpendControl` constructor and
`ProxyHandle` does not expose it, so both surfaces persist to
spending.json and the output instructs the operator to restart the
proxy/gateway (src/commands/policy.ts:45-46).

### Tests

- src/commands/policy.test.ts: argument rejection table asserts the store
  is never opened on invalid input (L98-116); nickname / malformed amount /
  malformed file all leave spending.json byte-identical (L167-190); a
  dropped write is reported as "Write did not land" (L118-129); a payee
  blocked through the command stops the signer before it is invoked (L31-63);
  plugin handler arg-splitting shares the same implementation (L131-139).
- src/proxy.spend-policy.test.ts now configures its blocked payee through
  `runPolicyCommand` instead of calling `setPolicy()` directly, so the
  existing `paidHits === 0` assertion pins the full path from operator
  surface to pre-sign hook.

### Notes

- `POLICY_LISTS` is exported from spend-control.ts (src/spend-control.ts:43)
  for the usage text and list-name validation.
- `KNOWN_NETWORKS` (src/commands/policy.ts:30) duplicates the set of payable
  chains; when a third chain is added, this list needs updating in the same
  change. Follow-up candidate: derive it from a single exported constant.
- Unstaged `package-lock.json` version bump (0.12.262 -> 0.12.263) is the
  fa0b146 release sync, not part of this change.

## Summary by CodeRabbit

- **New Features**
  - Added a local `policy` command for viewing and updating spending limits and counterparty allow/block lists.
  - Added `/policy` support to the plugin, including updates through the running proxy.
  - The command works without a running proxy and explains when a restart is required.
  - Policy changes support payable EVM and Solana networks and confirm successful persistence.

- **Bug Fixes**
  - Improved validation for malformed addresses, amounts, arguments, and policy files.
  - Prevented invalid or unsuccessful policy updates from modifying stored configuration.
  - Added clearer handling for conflicting concurrent policy changes.


VickyXAI (changes requested):
Careful work, and the parts that usually go wrong here are right. Enum coverage is exact — all four `SpendWindow` and all four `PolicyList` members handled, unknown values rejected rather than silently created. Amount parsing is properly fail-closed (`/^\d+(\.\d+)?$/` plus `Number.isFinite`; probed `"5abc"`, `"0"`, `"1e3"`, and 400 nines → `Infinity`, all rejected). `requireAuth: true` genuinely matches `exclude.ts`. The 12 lines in `spend-control.ts` only add an export and a read-only getter; no guard weakened.

Your fail-closed claim holds: no path in `runPolicyCommand` writes before validating, and `spending.json` is byte-identical after every rejection (probed).

Two things I'd want changed before this lands, both because it is a money guard.

### 1. `remove` of the last allowlist entry silently disables the guard

`spend-control.ts:394` gates on presence **and** length:

```ts
const payeePolicySet =
  (this.limits.blockedPayees && this.limits.blockedPayees.length > 0) ||
  (this.limits.allowedPayees && this.limits.allowedPayees.length > 0);
```

An emptied list routes to `clearPolicy()`, which deletes the key. So `allowedPayees` stops meaning "only these" and starts meaning "everything". Probed on disk:

```
policy set allowedPayees <A>     → check(payTo: <B>) = allowed:false
policy remove allowedPayees <A>  → prints "allowedPayees: (none)"
                                 → check(payTo: <B>) = allowed:true
```

An operation that reads as tightening turns the control off, and `policy.ts:164` renders it as a neutral `(none)` — indistinguishable from never-configured. Same for `allowedNetworks` and `allowedAssets`.

Either refuse (require an explicit `policy clear <list>`) or say it out loud: `allowedPayees is now unset — every payee is permitted`. Same treatment for `policy limit <window> clear`.

### 2. A `/policy` write inside the gateway does not affect the process that is signing

`proxy.ts:2324` builds `SpendControl` once and nothing re-reads `spending.json`; the plugin command opens its own instance (`policy.ts:176`). You do disclose this in `RESTART_NOTE`, but the headline line of the result is `blockedPayees: [...]` (`policy.ts:212`), so an operator blocking a draining payee mid-incident sees success while the live signer keeps paying it.

That is the wrong failure mode for the one command someone reaches for under pressure. Best fix is to have `startProxy` expose the live `SpendControl` (or a reload hook) and mutate that. Failing that, make the restart requirement the first line of the output for `blockedPayees` / `allowed*` writes, not a trailing note.

### Minor

- `policy.ts:202-210` — the limits write is an unlocked whole-object read-modify-write, and the landed-check only inspects the key it changed. Two interleaved commands drop a sibling key with no error (probed: `policy set blockedPayees <A>` then `policy limit daily 5` left disk as `{"daily":5}`, blocklist gone, command reported success). Widening the check to compare the full `getLimits()` against `{...before, [key]: expected}` would at least surface it; a `mutateLimits(fn)` that loads-applies-saves in one step would fix it.
- `policy.ts:177` + `spend-control.ts:698-700` — the same write rewrites `history` from the open-time snapshot, discarding spend the proxy recorded in between, which reopens the rolling hourly/daily windows. Probed. A history-preserving `saveLimits()` sibling to `saveHistory()` would do it.
- `policy.ts:6-7` — the comment says every argument is checked before the store is opened; the `remove` membership checks necessarily run after `openControl()`. Opening is read-only so nothing is at risk, only the wording.

`tsc` clean, 831 tests pass.

VickyXAI:

VickyXAI:
Correction to my note above — I had the mechanism wrong.

CI checks out the **merge ref** (your head merged with `main` as of trigger time), not the PR head. So you do *not* need to merge `main` in: any new push retriggers and picks up the fix automatically. What does **not** work is re-running the existing run — that replays the original merge ref and stays red.

#304 is the proof: it pushed a commit at 02:13 (after #300 merged at 02:09) and went green without ever containing `a58c2ae`. Sorry for the misdirection.

twzrd-sol:
Thanks for probing these rather than taking them — all five reproduced, and each is its own commit so the fix and its test can be read together.

**1. `remove` emptying an allow-list** — refused now. `nextListValue` rejects a `remove` that would leave `allowedPayees`/`allowedNetworks`/`allowedAssets` empty, naming what would become permitted and pointing at the explicit `policy clear <list>` (`src/commands/policy.ts:186`). `blockedPayees` may still be emptied by `remove`, since that reads as intended loosening. Every write that leaves a key unset — `clear`, `limit <window> clear`, the `blockedPayees` case — now leads with what is permitted or uncapped as a result (`unsetMeaning`, `policy.ts:45`), and the show output renders unset keys the same way instead of `(none)`. Test: `refuses to remove the last allow-list entry` asserts the list is untouched and `check()` still refuses the other payee.

**2. Gateway writes not reaching the signer** — took the first option. `index.ts` keeps the proxy's `SpendControl` in a module variable, re-created on every proxy restart, passes it to `startProxy` (`index.ts:960`) and hands `/policy` a getter for it (`index.ts:2234`). A gateway write mutates that instance and persists through it; the result says `Applied to the running proxy`. The fresh-from-disk reopen is kept for the landed-check only (`PolicyCommandOptions.liveControl`, `policy.ts:92`). The CLI is a separate process and cannot reach the live instance, so there `NOT applied to a running proxy — …` is the first line of every write's output (`policy.ts:75`). Where that collides with point 1 (a CLI write that unsets a guard) the restart line takes line 1 and the unset warning is line 2 — the mid-incident failure mode seemed the one to protect. Test: `a gateway write reaches the live signer with no reopen` asserts the very instance refuses afterwards; `a CLI write … leads with the restart requirement` pins the ordering.

**Minor — landed check** — widened to the whole object: disk-before-write is captured (`policy.ts:231`), the expected state is that plus the one change, and every key is compared; each drifted key is named. Your probe is the test verbatim (`reports a stale instance dropping a sibling key`): `set blockedPayees` then a stale instance's `limit daily 5` now fails with `blockedPayees (unset) vs [...]` instead of reporting success. I stopped short of `mutateLimits(fn)` — a live instance that re-read disk before every mutation would also adopt hand-edits made since startup, which is a policy change in itself and felt like a separate conversation.

**Minor — history snapshot** — `SpendControlStorage.saveLimits()` added as the mirror of `saveHistory()` (`src/spend-control.ts:192`, file store `:301`, in-memory `:332`); `SpendControl.save()` routes a limits-only change through it (`:733`), writing limits with whatever history disk holds, and then stops treating the limits as dirty so later history-only saves preserve on-disk limits again. Test: a stale instance's policy write keeps the record the proxy wrote in between (was wiped to 0 before).

**Minor — comment wording** — corrected: validation happens before anything is written; `remove` membership checks run after a read-only open.

`tsc` clean, lint clean, 838 tests. Each new test was confirmed red against the previous commit before the fix went in.

twzrd-sol:
Three more, from the bot pass on the last push — all three reproduced, each its own commit.

**`index.ts:960` (Major) — live instance claimed while no proxy runs.** Reproduced: the variable was assigned before `startProxy()` resolved and never cleared on stop/reset/failed start. The getter handed to `/policy` now gates on `activeProxyHandle`, which every stop and reset path already clears, so without a running proxy the command falls back to the on-disk write and leads with the restart requirement (`30b6062`).

**`policy.ts:121` (Minor) — `set blockedPayees not-an-address` succeeds.** Reproduced. Entries for `allowedPayees`/`blockedPayees`/`allowedAssets` must now be an EVM address or a Solana base58 id (32–44 chars), the two forms the proxy can be quoted; anything else is refused naming both shapes. I did not take the suggested EVM-only rule — Solana payees are legitimate here and `check()` matches them exactly, so base58 keeps its case (`8bf86f9`).

**`policy.ts:245` (Major) — concurrent writes.** Fair: the landed-check I added reported a dropped sibling only after the stale snapshot had replaced it. I said earlier I'd stopped short of `mutateLimits(fn)`; this is smaller than that and answers the actual defect. `saveLimits()` is now a compare-and-swap: it takes the limits the writer last read and refuses with `SpendPolicyConflictError` if storage no longer holds them, re-reading immediately before the atomic rename so the window is the write, not the command. `SpendControl` keeps that baseline from load and from every successful write; on conflict it adopts what landed and rethrows, so a live instance never keeps enforcing a snapshot that lost. The command reports the conflict and writes nothing. Your probe is the test verbatim and now leaves the blocklist intact; a retry from current state succeeds (`817aeef`). Cross-process interleaving inside that narrow window is stated as a remaining limit, not claimed solved.

`tsc` clean, lint clean, 842 tests; each new test confirmed red against the previous commit.

VickyXAI (approved):
Both blocking findings are properly fixed, and the compare-and-swap work goes beyond what I asked for. I verified each claim against the code rather than taking the commit messages.

**Emptying an allow-list is now refused, not silently permitted.** `policy.ts:198-204` rejects removing the last entry and names the consequence:

> removing the last allowedPayees entry would unset the list, and then every payee is permitted. Use "policy clear allowedPayees" if that is what you want

Making `unsetMeaning()` a shared helper and reusing it in the show output is better than the warning I suggested — the "loosest possible state" is now impossible to reach by accident and impossible to misread as "never configured". Correctly distinguishes allow-lists from `blockedPayees`, where emptying is safe.

**`/policy` now reaches the live signer**, and the detail I would have missed is right: `liveSpendControl` (`index.ts:786`) is gated on `activeProxyHandle`, so a failed, superseded or stopped proxy cannot let the command claim "applied to the running proxy". That is the failure mode that would have hurt — a false success while the live signer keeps paying a payee you just blocked.

**The compare-and-swap is sound, and honest about its own boundary.** `saveLimits(limits, expect)` re-reads immediately before the atomic rename and throws `SpendPolicyConflictError` rather than replacing a whole object from a stale copy; on conflict the instance adopts what landed, clears `limitsDirty`, and rethrows so the caller decides. The comment says the window is "the write itself, not the whole command" — that is accurate, it narrows a load-then-save race rather than eliminating it, and saying so is the right call for a file-backed store.

Two claims it rests on, both of which hold:

- `canonicalLimits` sorts keys and relies on list order being normalized — `setPolicy` does call `normalizePolicyValues` (`spend-control.ts:433`), so equal policies cannot false-conflict on ordering.
- The landed check is now whole-object: `const before = openControl().getLimits()` (`policy.ts:242`), "the baseline the landed-check compares every key against, not just the one being changed."

**The new address validation gets the case rules right**, which is easy to get wrong. `normalizePayee` (`spend-control.ts:69`) lowercases only what matches `EVM_ADDRESS` and leaves base58 untouched, and both the stored list (`normalizePolicyValues`) and the incoming quote (`check()`, `spend-control.ts:479`) go through it. So a checksummed `allowedPayees` entry still matches a lowercase `payTo`, while Solana addresses stay case-sensitive. `PAYABLE_NETWORKS` as one exported constant is the right shape for the network list too.

842 tests, typecheck and ESLint clean.

### Nothing blocking left. One housekeeping note

The branch is 10 commits behind `main` and predates #300, so its own `CHANGELOG.md` still fails `prettier --check .` locally. Your CI is green because it builds the merge ref against current `main`, and merging resolves it — not a defect in this PR, just worth knowing if you run the check locally and see it fail.

The three minor items from my earlier review are all addressed by `7814110` and `817aeef` (whole-object landed check, history no longer carried from a stale snapshot, unlocked read-modify-write replaced by CAS). Good work on this one.

PR #304, opened 2026-09-02
fix(spend-policy): gate every signing path on SpendControl, not just the proxy

### LINKED ISSUE #302: Counterparty policy does not reach Polymarket fund/order/withdraw/redeem signing or `doctor`'s x402 client
Follow-up to #268, per the scope note there.

**Enforcement today (v0.12.263, `fa0b1468`): one site.** `src/proxy.ts:2323-2325` builds the proxy `x402Client` and calls `registerSpendPolicyHook` before `registerExactEvmScheme`. No other file calls `registerSpendPolicyHook` or `SpendControl.check()`.

**Paths that sign with the same wallet and never consult it:**

1. `src/doctor.ts:409-439` — a second `x402Client` (L409), EVM/SVM schemes (L410/L426), `wrapFetchWithPayment` (L435), paid `/v1/chat/completions` (L438-439) at ~$0.003–$0.01/run (`DOCTOR_MODELS` L370-379). The proxy's shape minus one line.
2. `src/polymarket/fund.ts:122` — `createPaymentPayload()` (`@blockrun/llm`) signs an EIP-3009 authorization to `bridge` on `eip155:8453`; L132-133 pays the $0.01 fee via `new BlockrunClient({ privateKey })`. No `SpendControl` import.
3. `src/polymarket/orders.ts:433` — `clob.createAndPostOrder()`. `confirm:true` (L407), `POLYMARKET_MAX_BET_USD` (L377) and `reserveBet` (L429) are per-order notional caps on a separate ledger, not a counterparty check.
4. `src/polymarket/withdraw.ts:147` and `src/polymarket/redeem.ts:155` — `sendTransaction` / `sendWalletBatch`, same shape.

Net: an `allowedPayees` list that makes the proxy refuse every other payee does nothing here. `docs/configuration.md` reads wallet-level; enforcement is proxy-level.

**Fix**
- `doctor.ts`: `registerSpendPolicyHook(x402, new SpendControl())` before L410 — identical to `proxy.ts:2325`.
- Polymarket paths: `control.check(amount, { payTo, network, asset })` before the signer — CAIP-2 network, contract addresses from `constants.ts`; `!allowed` throws `SpendPolicyError`, nothing is signed. The fee leg inside `BlockrunClient` needs its own hook or a doc note — `@blockrun/llm` 3.13.2 exposes none.
- Fail closed: no parseable amount with limits configured → refuse, as `assertSpendPolicyAllows` already does.

**Tests** — wiring-level, mutation-checked like `proxy.spend-policy.test.ts`: in-memory `SpendControl` with `blockedPayees` matching the counterparty; assert the signer is never called on each path. Deleting each check line must turn the test red.

PR to follow.

twzrd-sol:
Closes #302.

`registerSpendPolicyHook` has exactly one call site: `src/proxy.ts:2325`, registered before `registerExactEvmScheme` on the proxy's `x402Client`. Five other places sign with the same wallet and never consulted it — a strict `allowedPayees`/`allowedNetworks` list refused every payee on the proxy while doing nothing here.

## What's covered now

| Path | Check placed before | Counterparty handed to policy |
|---|---|---|
| `src/doctor.ts` | `registerExactEvmScheme` (own `x402Client`, doctor's only one) | whatever the 402 quotes — identical to the proxy |
| `src/polymarket/fund.ts` | `createPaymentPayload` | payTo=bridge, `eip155:8453`, Base USDC, exact `amountMicro` |
| `src/polymarket/orders.ts` | `clob.createAndPostOrder` / `createAndPostMarketOrder` | payTo=CTF or NegRisk exchange by market, `eip155:137`, pUSD, notional |
| `src/polymarket/withdraw.ts` | `sendWalletBatch` / `wallet.sendTransaction` (both branches) | payTo=**recipient** (agent-chosen `to_address`, not the one-time bridge), `eip155:8453`, Base USDC, exact pUSD micros |
| `src/polymarket/redeem.ts` | `sendWalletBatch` / `wallet.sendTransaction` (both branches) | payTo=ConditionalTokens or NegRiskAdapter by market, `eip155:137`, pUSD, amount 0 |

Redeem moves no capital out — it burns the wallet's own outcome tokens and credits collateral back to the same wallet, so the checked amount is 0 and the change is a no-op for every operator with nothing configured. It's covered anyway because `allowedPayees`/`allowedNetworks` are total statements about what the agent may sign, and an unwrapped path makes that silently non-total; the cost falls only on an operator who wants a strict allowlist, who now lists two more contracts — the same class of cost `orders.ts` already imposes.

Shared helper `src/polymarket/spend-policy.ts`: one `SpendControl` per process so session windows span tool calls the way the proxy's do; refusal is `SpendPolicyError`, thrown before the signing function runs; the reservation settles when the signer returns and releases if it throws, so a failed submit doesn't consume budget. `mapClobError` gets a 3-line `instanceof` branch so the agent sees "Refused by spend policy" instead of a raw CLOB error.

## Known limitations

**The two `SpendControl` instances don't share runtime state — only the on-disk file.** `proxy.ts:2324` and `polymarket/spend-policy.ts:30` each construct their own instance against the same `spending.json`. Each loads history once at construction, so after that, `hourly`/`daily`/`session` windows are enforced **per-surface** — LLM spend is invisible to a Polymarket check and vice versa, beyond whatever was on disk when each instance started. Worse, each instance's `saveHistory()` rewrites the file's history with only its own view, so on a process restart the two surfaces' histories can clobber each other — last-writer-wins, and the on-disk audit trail degrades rather than accumulating cleanly. The fix is a single process-wide `SpendControl`, threaded through the existing `PolymarketSpendDeps` seam from wherever the proxy and the Polymarket tool are wired together in `index.ts`. Filing as a fast follow-up rather than blocking this PR on it — the policy is still enforced correctly per-call, just not with a perfectly unified ledger yet.

**A market sell reserves budget too.** `orders.ts` reserves `size × bid` before submitting regardless of side, so a buy-then-sell round trip double-counts against the amount windows — heavy LLM spend could, in principle, lock an agent out of *exiting* a Polymarket position through its own cap. This is the fail-closed default and probably the right one, but it's a real behavior change worth a changelog line, not just new coverage.

**`orders.ts` now makes Polymarket notional consume the same amount-window ledger as LLM spend.** The file's header used to say bets were "deliberately NOT the x402 budget ledger" — that comment is updated. An operator with `perRequest: 0.10` set for LLM calls will now see a default $25 order refused by the same cap. With nothing configured in `spending.json`, behavior is unchanged.

**The $0.01 fee leg inside `fund.ts`'s `BlockrunClient.post`** signs internally with no pre-sign hook exposed by `@blockrun/llm` (checked its `.d.ts` — no hook option). It only runs after the principal payment has passed policy, and its payee is fixed to BlockRun's own gateway, so the exposure is narrow. No workaround attempted; flagging rather than hiding it.

## Tests

Wiring-level, matching `proxy.spend-policy.test.ts`'s existing style: in-memory `SpendControl` with the target counterparty on a blocked/denied list, assert the signing function is never called. Every new check was confirmed red before its fix and green after (each temporarily removed and restored, verified by grep). `redeem.ts`'s tests additionally cover the neg-risk routing case — a `ConditionalTokens`-only allowlist correctly refuses a neg-risk market's redeem, which resolves to `NegRiskAdapter`.

74 test files / 820 tests pass. Typecheck and lint clean.

## Summary by CodeRabbit

* **New Features**
  * Added spend-policy checks before signing AI-assisted payments, funding, trades, withdrawals, and redemptions.
  * Transactions now validate destinations, networks, assets, and amounts before submission.
  * Added clearer refusal messages for blocked spending.

* **Bug Fixes**
  * Improved confirmation handling for direct wallet transfers and redemptions.
  * Ensured failed order submissions release spend reservations.

* **Tests**
  * Added coverage for blocked and allowed transactions, alternate trading routes, and AI-assisted payments.

### DIFF
--- a/src/doctor.spend-policy.test.ts
+++ b/src/doctor.spend-policy.test.ts
@@ -1,0 +1,84 @@
+import { createServer, type IncomingMessage, type Server, type ServerResponse } from "node:http";
+import type { AddressInfo } from "node:net";
+import { afterAll, beforeAll, describe, expect, it } from "vitest";
+import { generatePrivateKey } from "viem/accounts";
+import { wrapFetchWithPayment } from "@x402/fetch";
+
+import { createDoctorX402Client } from "./doctor.js";
+import { SpendControl, InMemorySpendControlStorage, CAIP2_BASE } from "./spend-control.js";
+
+/**
+ * Doctor builds its own x402 client and pays a real endpoint. Until this
+ * landed it registered the signing schemes but not the spend-policy hook, so a
+ * blocked payee the proxy refused was still paid by `clawrouter doctor`.
+ * Same shape as proxy.spend-policy.test.ts: a mock upstream that challenges
+ * with a 402 naming a blocked payee, and a count of retries that carried a
+ * signed payment. That count must stay at zero.
+ */
+describe("doctor's x402 client enforces spend policy before signing", () => {
+  const blockedPayee = "0xAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAaAa";
+  let upstream: Server;
+  let url: string;
+  let paidHits = 0;
+  let unpaidHits = 0;
+
+  beforeAll(async () => {
+    upstream = createServer((req: IncomingMessage, res: ServerResponse) => {
+      req.on("data", () => {});
+      req.on("end", () => {
+        if (req.headers["x-payment"] || req.headers["payment-signature"]) {
+          paidHits++;
+          res.writeHead(200, { "Content-Type": "application/json" });
+          res.end(JSON.stringify({ choices: [{ message: { content: "paid" } }] }));
+          return;
+        }
+        unpaidHits++;
+        const challenge = {
+          x402Version: 2,
+          resource: { url: "http://127.0.0.1/v1/chat/completions" },
+          accepts: [
+            {
+              scheme: "exact",
+              network: CAIP2_BASE,
+              amount: "10000",
+              asset: "0x833589fCD6eDb6E08f4c7C32D4f71b54bdA02913",
+              payTo: blockedPayee,
+              maxTimeoutSeconds: 60,
+              extra: {},
+            },
+          ],
+        };
+        res.writeHead(402, {
+          "Content-Type": "application/json",
+          "PAYMENT-REQUIRED": Buffer.from(JSON.stringify(challenge)).toString("base64"),
+        });
+        res.end(JSON.stringify({ error: "payment required" }));
+      });
+    });
+    await new Promise<void>((resolve) => upstream.listen(0, "127.0.0.1", resolve));
+    url = `http://127.0.0.1:${(upstream.address() as AddressInfo).port}/v1/chat/completions`;
+  });
+
+  afterAll(async () => {
+    upstream.closeAllConnections?.();
+    await new Promise<void>((resolve) => upstream.close(() => resolve()));
+  });
+
+  it("refuses a blocked payee and never attaches a payment", async () => {
+    const control = new SpendControl({ storage: new InMemorySpendControlStorage() });
+    control.setPolicy("blockedPayees", [blockedPayee]);
+    const x402 = createDoctorX402Client({ walletKey: generatePrivateKey(), spendControl: control });
+    const paymentFetch = wrapFetchWithPayment(fetch, x402);
+
+    await expect(
+      paymentFetch(url, {
+        method: "POST",
+        headers: { "Content-Type": "application/json" },
+        body: "{}",
+      }),
+    ).rejects.toThrow(/blocked by policy/i);
+
+    expect(unpaidHits).toBeGreaterThan(0);
+    expect(paidHits).toBe(0);
+  }, 20_000);
+});
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -23,6 +23,7 @@ import { BalanceMonitor } from "./balance.js";
 import { getSolanaAddress } from "./wallet.js";
 import { getStats } from "./stats.js";
 import { getProxyPort } from "./proxy.js";
+import { registerSpendPolicyHook, SpendControl } from "./spend-control.js";
 import { VERSION } from "./version.js";
 
 // Types
@@ -380,6 +381,27 @@ const DOCTOR_MODELS: Record<DoctorModel, { id: string; name: string; cost: strin
   },
 };
 
+/**
+ * Build the x402 client doctor pays with. The spend-policy hook is registered
+ * here, before any signing scheme, exactly as startProxy does: doctor's paid
+ * call (~$0.003-$0.01) must not be a way around the operator's allow/deny
+ * lists. This is the only place doctor constructs an x402Client, so the test
+ * that drives it pins the wiring, not just the helper.
+ */
+export function createDoctorX402Client(opts: {
+  walletKey: string;
+  /** Default: the same on-disk policy the proxy reads. Inject in tests. */
+  spendControl?: SpendControl;
+}): x402Client {
+  const account = privateKeyToAccount(opts.walletKey as `0x${string}`);
+  const publicClient = createPublicClient({ chain: base, transport: http() });
+  const evmSigner = toClientEvmSigner(account, publicClient);
+  const x402 = new x402Client();
+  registerSpendPolicyHook(x402, opts.spendControl ?? new SpendControl());
+  registerExactEvmScheme(x402, { signer: evmSigner });
+  return x402;
+}
+
 // Send to AI for analysis
 async function analyzeWithAI(
   diagnostics: DiagnosticResult,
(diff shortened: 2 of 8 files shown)


VickyXAI (commented):
This is the strongest PR in the queue right now. It closes a real gap — `registerSpendPolicyHook` had exactly one call site while five other paths signed with the same wallet and ignored the operator's lists.

I verified the load-bearing claim rather than taking it. `CANONICAL_AMOUNT = /^\d+$/` plus `assertSpendPolicyAllows`:

```
5    -> "5000000"  -> parsed 5        NaN  -> "NaN"       -> REJECTED
0    -> "0"        -> parsed 0        -5   -> "-5000000"  -> REJECTED
1e15 -> "1e+21"    -> REJECTED        Inf  -> "Infinity"  -> REJECTED
```

Every non-canonical form is refused, so the fail-closed behaviour is exactly as documented.

I also enumerated every signing site in the repo rather than trusting the coverage table. Two apparent gaps resolve clean: the hook attaches to the `x402Client` via `onBeforePaymentCreation`, so doctor's later `registerExactSvmScheme` is covered, and `payment-preauth.ts` is covered transitively because it takes the caller's client. Worth stating explicitly since it is not obvious from the table.

The withdraw gate is the best decision in the PR — policy sees the destination leg (agent-chosen `to_address`, Base, USDC) rather than the one-time bridge address that cannot be allowlisted, with both signing branches inside the closure. That is the right call and the comment explains it well.

`tsc` clean, 820 tests pass.

Two things to resolve, neither blocking.

### 1. `setup.ts:303` signs and is not gated

```ts
const res = await sendWalletBatch(calls, depositWallet, "Approval batch");
```

`buildApprovalCalls` emits ERC-20 `approve` and ERC-1155 `setApprovalForAll`. Approvals do not move capital but grant the authority to — arguably the strongest thing this wallet signs. Risk is bounded (targets come from `readApprovals`, not agent input, and it is `confirm`-gated), so I am not asking you to gate it necessarily. But the module comment says "Every Polymarket path that signs something that moves the user's capital", and this one is neither covered nor listed as deliberately excluded. Either gate it or name it in the table as out of scope and why.

### 2. The two ledgers now disagree on a rejected order

A CLOB rejection resolves with `{success:false}` rather than throwing, so `signUnderSpendPolicy` settles the SpendControl reservation. Then:

```ts
const placed = r?.success !== false && (r?.orderID || r?.status === "matched");
if (!placed) {
  releaseBet(notional, input.agent_id);
  throw new Error(r?.errorMsg || "order rejected");
}
```

The legacy ledger is refunded, the SpendControl window is charged. Two caps gating the same wallet now account for the same event differently.

"Signed = spend" is defensible and consistent with the x402 hook's stated conservatism — the wallet did sign before the CLOB rejected. But right now the choice is implicit and the two disagree by accident. Pick one and say so in the comment; I lean toward keeping settle-on-sign and making `releaseBet` match.

### Note

Checked this for vendor insertion given the history on the pre-sign path. Clean — every new address is a test fixture (`0xAaAa…`, `0x1111…`, `0xdead…`) plus canonical Base USDC. No branding, no attribution, no seat.

Happy to approve once the two points above are answered. Note `Lint & Typecheck` is red through no fault of yours — a CHANGELOG formatting regression I introduced on `main`; #300 fixes it.

VickyXAI:

VickyXAI (commented):
Both points addressed, and I verified the fixes rather than taking them.

**The two-ledger disagreement is properly closed.** Moving the `placed` check inside the `signUnderSpendPolicy` callback and throwing there is the right shape — a resolved-but-rejected CLOB response now propagates as a throw, so `signUnderSpendPolicy` releases the reservation *and* the outer catch releases the session bet. Both ledgers now agree on a rejected order. You went the opposite way from my lean (release-both rather than settle-both), and I think your call is the better one: the wallet signed, but no capital moved and the order never rested, so charging the window would overstate spend. The comment says which semantic was chosen and why, which was the actual ask.

**The redeem receipt fix is correct, and I checked the claim it rests on.** `sendWalletBatch` does `const confirmed = await response.wait()` (`relayer.ts:163`), so the relayer path genuinely confirms internally and `if (eoa)` is right. Worth noting for the record that `main` already waited unconditionally inside the EOA branch (`redeem.ts:156`), the restructure in the earlier commit inverted it to `if (!eoa)`, and `0b03d68` restores correct behaviour. Net vs `main` is right. Self-caught, but it is a good illustration that hoisting a send into a closure moves the receipt wait out of its guard — worth a second look anywhere else the same restructure was applied.

`tsc` clean, 823 tests pass, and `Lint & Typecheck` is green now that your push recomputed the merge ref against the fixed `main`.

### Still open from my earlier review

`setup.ts:303` — `sendWalletBatch(calls, depositWallet, "Approval batch")` still signs ungated, and neither the module comment nor the PR body mentions it. I am not asking you to gate it: targets come from `readApprovals`, not agent input, and it is `confirm`-gated, so the exposure is bounded. But `spend-policy.ts` opens with "Every Polymarket path that signs something that moves the user's capital", and an unlimited ERC-20 `approve` / `setApprovalForAll` is the strongest authority this wallet grants. One line in the coverage table naming it as deliberately out of scope, with the reason, would close it.

Happy to approve once that is either gated or named.

VickyXAI (approved):
Both review points addressed and verified, and I pushed the last one myself rather than make you round-trip for a comment (`65696e6`).

Recorded there that the `setup.ts` approval batch stays ungated **deliberately** — gating it would be actively harmful, since the spenders are Polymarket's own exchange contracts and an operator running a tight `allowedPayees` list naming only their payout addresses would have setup refused. It is already bounded: targets come from `readApprovals()`, never agent input, and the batch is confirm-gated behind an explicit preview.

Good work on this one. The withdraw gate seeing the destination leg rather than the one-time bridge address is the right call, and the rejected-order fix landed on the better of the two semantics.

PR #313, opened 2026-09-02
feat: add ClawRouter Desktop and isolate OpenClaw plugin id

KillerQueen-Z:
## Summary

- add the macOS ClawRouter Desktop control plane for OpenClaw, Codex, Hermes, DeepSeek Harness, and Pi, including wallet chain switching, balances, model pricing/context, and Coinbase Onramp
- rename the OpenClaw plugin id from `clawrouter` to `blockrun-clawrouter` while keeping the npm package and CLI stable
- migrate only proven legacy BlockRun settings and preserve OpenClaw’s bundled `clawrouter` plugin, third-party plugins, wallet, credentials, and config
- harden setup/update/reinstall/uninstall rollback across extension and OpenClaw 2 npm-project layouts
- require authenticated proxy identity and spawned-process port ownership before Desktop configures an agent

## Validation

- core: 870 passed, 1 skipped; typecheck and ESLint passed
- integration: 23 passed, 3 skipped
- Desktop: 33 passed, 1 skipped; typecheck and production build passed
- packaged and launched the arm64 macOS app; transparent 2048px icon verified
- OpenClaw 2026.8.2: official `clawrouter` and `blockrun-clawrouter` loaded together with no diagnostics; config validation, install/setup, uninstall/reinstall, and failed-install rollback passed
- real free-model request through the packaged ClawRouter dist returned `OK`
- exact npm pack smoke check passed
- production audit: no high/critical advisories; Desktop app has 0 advisories (core retains upstream low-only Polymarket/ethers advisories with no available fix)

Ready for review.

## Summary by CodeRabbit

* **New Features**
  * Added a macOS ClawRouter Desktop app with agent connections, model discovery, usage views, settings, wallet management, funding, and theme support.
  * Added integrations for Codex, DeepSeek Harness, Hermes, OpenClaw, and Pi.
  * Added shared BlockRun Core wallet support for Base and Solana, with non-destructive legacy migration.
  * Added richer model metadata and the `blockrun_polymarket` tool contract.
  * Added the `clawrouter policy` command.

* **Documentation**
  * Added Desktop setup, wallet, configuration, coexistence, and troubleshooting guidance.

* **Security**
  * Improved authentication, rollback, input validation, and private file handling.


VickyXAI (changes requested):
Reviewed against `main` @ `9ebc22f` (v0.12.265). Every check the description claims reproduces: 870 passed / 1 skipped, `npm run typecheck`, `npm run lint` and `prettier --check .` all clean.

Big picture first: **this is a substantial improvement on #291.** Four of that PR's six blocking findings are gone outright, and the plugin-config work is better than what I shipped — details below, including two bugs of mine it corrects. Two findings survive unchanged, one of them worse than it was, and the new surface brings three of its own.

---

## What got fixed, and credit where it is due

**The wallet rewrite is gone.** `src/auth.ts` is not in the diff and precedence is back to `saved → env → generate`. That removes #291's findings 1-4 in one move: the silent payer flip, the Solana fund-vs-spend address mismatch, the startup brick on a malformed session file, and the `/wallet export` mnemonic that could not recover the wallet in use.

**`src/openclaw-plugin-config.ts` corrects two real bugs in what I shipped in v0.12.265 / #308.** I want this on the record because it was my code:

1. My migration *deleted* `plugins.entries.clawrouter`. After the rename that key belongs to **OpenClaw's bundled plugin**, so I was silently mutating a different product's config. This PR preserves the entry and moves only BlockRun-owned fields (`walletKey`, `routing`, and the same under `config`). That is correct and I was wrong.
2. My migration *renamed* `plugins.installs.clawrouter` to the new id. OpenClaw's own docs, in the 2026.8.2 tarball: *"`plugins.installs` is a retired authored-config surface… Run `openclaw doctor --fix` to import legacy config records into the index and remove the retired key."* Deleting it (`stripUnsupportedInstalls`) is right; renaming it writes into a key upstream wants gone.

Mine also ran unconditionally on every gateway start. Gating on `explicitSetup` / `legacyBlockRunInstall` is the safer design.

The `isGatewayMode()` guard on the `openclaw.json` write survived the refactor (`src/index.ts:438`), so there is no install-time `baseHash` rollback risk.

---

## Blocking

### 1. `/admin/models` does not exist, so the hardcoded catalog wins on every launch

`apps/desktop/electron/core/manager.ts:378-386` prefers a live endpoint and falls back to the bundle:

```ts
const detailed = await fetchJson(`${root}/admin/models`, fetcher);
return Array.isArray(detailed.value?.data) ? detailed : fetchJson(`${root}/v1/models`, fetcher);
```

That endpoint is not on `main` and is not added here — `git grep "admin/models" -- src` is empty on both. #291 *did* add it (`buildDetailedModelList` in `src/proxy.ts`); this rewrite dropped the endpoint and kept the caller. This PR's own tests hardcode it as 404 (`apps/desktop/tests/manager.test.ts:149`).

So it always falls to `/v1/models`, whose payload is four fields (`src/proxy.ts:1555-1560`): `id`, `object`, `created`, `owned_by`. No name, price, context window or capability flags. Every `live ?? bundled` in `manager.ts:249-258` and `App.tsx:1234-1242` therefore resolves to the bundle, on every field, on every launch. "Prefer live, fall back to bundle" is real in code and dead in practice.

### 2. The bundled catalog is byte-identical to #291 and badly stale

`git rev-parse pr-291:…/model-catalog.ts pr-313:…/model-catalog.ts` → same blob `b4c4128…`. Zero refresh, despite the file's own header saying "Regenerate whenever the root model catalog changes."

Re-measured against current `main`: **197 field disagreements across 110 ids**; 86 registry ids missing; 16 ids with wrong prices, nearly all understating cost:

| id | bundled in/out | registry in/out |
|---|---|---|
| `deepseek/deepseek-v4-pro` | 0.435 / 0.87 | **1.32 / 3.96** |
| `fable`, `fable-5`, `fable-5.0` | 5 / 25 | **10 / 50** |
| `gemini-3.5-flash` | 0.5 / 3 | **1.5 / 9** |
| `zai/glm-5` | 0.6 / 1.92 | **1 / 3.2** |
| `grok` | 1.5 / 4 | **2.5 / 9** |

Missing entirely: `claude-opus-5`, `claude-sonnet-5`, `claude-fable-5`, all eight `openai/gpt-5.6-*`, `zai/glm-5.3`, `tencent/hy3`, `google/gemini-3.6-flash`, and more.

**Free tier is the worst case: `FREE_MODELS` (`src/proxy.ts:180-188`) lists 7 and the bundle carries 1** — missing `free/nemotron-3.5-lightning`, which is what the `free` alias resolves to.

It also reverts `fea53e8 fix(models): gemini-2.5-flash-lite reads images` — `model-catalog.ts:437-448` has `vision: false` where `src/models.ts:1238-1247` has `true`.

`adapters/pi.ts` makes this durable rather than cosmetic: it never calls `fetchCatalog`, hits `/v1/models` directly (`pi.ts:80`), resolves price/context/vision entirely from the bundle (`pi.ts:98-114`), and writes them into the user's real `~/.pi/agent/models.json` (`config.ts:105`). `vision` decides `input: ["text","image"]` vs `["text"]`, so `gemini-2.5-flash-lite` is persisted as text-only.

No generator, no test, no CI check — `git grep -l model-catalog` returns three consumers and the file. Compare `src/router/chain-models-in-catalog.test.ts`, which exists precisely because an uncarried id is a cost-cap hole.

**Ask:** delete the bundle and require the live endpoint — which means restoring `/admin/models` from #291. If a bundle must stay, it needs a generator plus a test that fails when it diverges.

### 3. The install scripts downgrade `openclaw.json` from 0600 to 0644

`scripts/reinstall.sh:726-729` (and `:366`, `:687`, `:716`, `:812`, plus `scripts/update.sh`):

```js
function atomicWrite(filePath, data) {
  const tmpPath = filePath + '.tmp.' + process.pid;
  fs.writeFileSync(tmpPath, data);
  fs.renameSync(tmpPath, filePath);
}
```

`writeFileSync` creates the temp file at the default mode, and the rename carries it over. Probed on macOS: `before: 600 → after: 644`. My `~/.openclaw/openclaw.json` is `-rw-------` today.

That file can hold a wallet private key: `openclaw.plugin.json` declares `configSchema.properties.walletKey` ("EVM wallet private key (0x…)") with `uiHints.walletKey.sensitive: true`, and these very scripts move `walletKey` between entries (`reinstall.sh:251,745,751`). `src/cli.ts:892` gets it right with `{ mode: 0o600 }`.

**Fix:** pass `{ mode: 0o600 }` to every `writeFileSync(tmpPath, …)` in the embedded helpers, or `chmodSync` after the rename.

---

## Worth fixing, not blocking

**Upgrade without `setup` gets no migration.** `prepareBlockRunPluginConfig` is called only from `src/cli.ts` (`:743`, `:761`, `:833`) — never from `src/index.ts`, where the old version ran on gateway start. A user who runs `npm update -g` and restarts the gateway is not migrated, so with `plugins.allow` set, `blockrun-clawrouter` is absent from an exclusive allowlist and the plugin cannot load. Removing the *unconditional* migration was right; the fix is a guarded one on the activation path too, not a restoration of mine.

**Packaging is not reproducible.** `stage-root-runtime.mjs` is deleted and `dist` no longer installs or stages the runtime, but `build.extraResources` still ships `runtime/node_modules/**/*` — a path that is gitignored and that no script populates. A clean checkout packages an empty runtime; a machine with leftovers packages whatever it has. And `runtime/pnpm-lock.yaml` still pins `@blockrun/clawrouter@0.12.212`, which I confirmed declares plugin id **`clawrouter`** — so if that directory is present at package time, the shipped app carries a ClawRouter that re-introduces the exact collision this PR fixes.

**Port ownership proves less than it looks.** `supervisor.ts:133-155` asserts *a* listener on the port descends from the spawned child, not that the socket answering `127.0.0.1` is that child's. Port 8402 is safe only because `src/proxy.ts:3593` binds `127.0.0.1` explicitly; 8403 is served by the out-of-repo Codex bridge, and if that binds wildcard a local process can shadow it while `lsof` still lists the genuine child. `isModelService` (`supervisor.ts:158-167`) only checks that `data` is an array, which is trivially forged, and `codex.ts:42` then writes that `base_url` into `~/.codex/config.toml`.

**Rollback reports success it cannot know.** `files.ts:124-147` restores in a bare loop; a throw on file *n* leaves *n..end* un-restored, and `manager.ts:96-103` returns `rolledBack: true` unconditionally. For `dsh` that is `settings.yaml` restored with `.credentials.yaml` not — a mismatched pair reported as a clean rollback.

**Unpinned installs with lifecycle scripts.** `runtime.ts:69-79` runs `npm install <pkg>@latest`; only `pi.ts:43-45` passes `ignoreScripts`. `dsh.ts:40` and both supervisor calls do not, and `openclaw.ts:63` shells `npx -y @blockrun/clawrouter@latest setup`. One Connect click runs whatever the current release's `postinstall` contains.

**Backups accumulate.** `src/cli.ts:617-620` writes `openclaw.json.before-blockrun-clawrouter.<ts>` on every `setup` and never deletes it; `reinstall.sh:174-177` copies the raw private key to `wallet.key.bak.<ts>` and `cleanup_backups` (`:42-55`) removes every other backup but not that one.

---

## Electron hardening: still clean

Re-verified rather than assumed: `contextIsolation: true` / `nodeIntegration: false` / `sandbox: true` (`main.ts:33-38`), `will-navigate` and `setWindowOpenHandler` guarded (`:44-47`), CSP in `index.html:7-10`, every IPC handler behind `requireTrustedRenderer` plus a parser, `preload.ts` exposing seven typed invokes, `process.ts:27` spawning without `shell: true`, `files.ts:40-51` writing 0600 with a symlink-cycle guard, and no key or mnemonic reaching the renderer (`manager.ts:298-308` reads `~/.blockrun/.session` but returns only the derived address). All three regressions above are in the new surface, not the parts #291 cleared.

---

Also: #291 is still open and this supersedes it. Worth closing that one with a pointer here so the review history does not fork.

VickyXAI (commented):
Reviewed `23fc5d4` + `30eb607`. The balance change is correct — I verified the part that worried me rather than assuming — but it does not touch any of the three blocking findings, so this stays as-is.

**The new local derivation is correct.** `manager.ts:522-542` re-derives the Solana address from the mnemonic (SLIP-10, `m/44'/501'/0'/0'`) with `createHmac` and a hand-rolled `base58Encode`. My concern was that a second derivation would disagree with the one the proxy actually signs with — which is exactly #291's finding 2, where the UI told users to fund an address the proxy never spent from.

It does not. I ran both implementations against three mnemonics:

```
MATCH true | HAgk14JpMQLgt6rVgv7cBQFJWFto5Dqxi472uT3DKpqk
MATCH true | BLeUXTx9thHGT7VJUtF9vHEmfMDgW1nnKZ9UVer2CoLX
MATCH true | E48cosDiQZK1iDSsyUzhvW4WxJeoKuDk5qgcdkmANV4N
```

Byte-identical to `deriveSolanaKeyBytes` + `getSolanaAddress` from `src/wallet.ts`. The `base58Encode` leading-zero case is handled correctly too (`for (const byte of value) { if (byte !== 0) break; encoded = \`1${encoded}\` }`), which is the usual way a hand-rolled encoder goes wrong. And the mnemonic stays in the main process — `localWallets()` returns addresses only, so the "no key material reaches the renderer" property from the earlier review still holds.

**What I would still change: nothing enforces that they keep agreeing.** This is now a second implementation of wallet derivation, in a second package, with no test comparing it to `src/wallet.ts`. Today they match. If the canonical derivation ever changes — path, SLIP-10 details, encoding — the Desktop silently diverges and starts displaying an address the proxy does not pay from, and the failure is invisible until someone funds the wrong one.

Cheapest fix is a test in `apps/desktop/tests/` that asserts the two produce the same address for a fixed mnemonic. Better is not to duplicate at all: the Desktop already ships `@blockrun/clawrouter` in `runtime/`, so it can consume the exported `deriveSolanaKeyBytes` rather than reimplement it.

**Still open from the review above** — none of these are addressed by these two commits:

1. `/admin/models` does not exist, so the hardcoded catalog wins on every launch
2. that catalog is byte-identical to #291's and carries 197 field disagreements, 16 wrong prices, and 1 of 7 live free models
3. `scripts/reinstall.sh` downgrades `openclaw.json` from 0600 to 0644, and that file can hold a wallet private key

KillerQueen-Z:
@VickyXAI Thanks for the detailed review. I pushed fdace18, which addresses all three blockers and the related maintenance gaps:

- `/v1/models` now exposes authoritative pricing/context/capability metadata; Desktop and Pi consume it directly, and the 2,441-line stale bundled catalog is removed.
- Installer/config/auth writes preserve `0600`; wallet adoption refuses backup collisions; docs no longer print `.session`.
- Runtime installs are frozen and version-pinned, CLI preflights are bounded, balance RPCs are cached with stale-on-error fallback, Solana derivation is shared, read-only wallet commands no longer create wallets, and rollback reporting now distinguishes actual rollback from preflight failure.

Validation: root 896 passed + 1 skipped; Desktop 45 passed + 1 skipped; typecheck/lint/format/build all pass; macOS ARM app packaging succeeds and bundled DSH/Codex/Pi executables were smoke-tested. Could you please re-review the latest commit?

VickyXAI (approved):
All three blockers verified fixed against `fdace18`, by reading the code rather than the commit message:

- `model-catalog.ts` is gone from the tree, and `/v1/models` now carries `name`/`context_window`/`max_output`/prices/`reasoning`/`vision`/`agentic`/`tool_calling` from the live registry (`proxy.ts:1583-1615`). I measured the new alias→registry join across all 287 unique `OPENCLAW_MODELS` entries: 0 misses, 0 vision disagreements.
- Every `atomicWrite` in `reinstall.sh`/`update.sh` passes `{ mode: 0o600 }` and re-`chmod`s after the rename.
- Runtime repinned to 0.12.265 (was 0.12.212, the old plugin id), `stage:runtime` restored to `dist`, installs pinned with `--ignore-scripts`.

I pushed `ad9fb43` rather than send you back for two small things:

1. The gateway-start plugin-id migration. v0.12.265 ran it on every gateway start; this branch left it only on `setup`/`update`/`reinstall`, so `npm update -g` + restart went unmigrated. It now runs from `injectModelsConfig`, where the write is already `isGatewayMode()`-gated, and only when BlockRun-owned fields prove the legacy entry is ours.
2. `runtime-version.test.ts` pinned the constant to the root package but not to the lockfile that stages the runtime — a version bump would have dropped Connect into a live `npm install`.

One follow-up, filed separately, not blocking: with a pre-existing `~/.blockrun/.session` written by another BlockRun product, Core outranks a funded legacy `wallet.key` and payment moves wallets on upgrade. Which side should win is your call, and it interacts with Desktop reading Core directly.

Nice work on the catalog removal in particular — deleting the bundle was the right answer over regenerating it.

PR #314, opened 2026-09-02
chore(brand): re-apply blockrun#469's canonical script

VickyXAI:
Reverses #310, which was **correct when it landed** and is now the last blocker.

## What changed since #310

At 14:41 UTC ClawRouter was the only consumer carrying the 360-line script, so blockrun main compared its own 270-line canonical against 15 copies and ClawRouter was the lone mismatch. #310's reasoning — "merging #469 moves the failure from 1 repo to the other 14" — was accurate at that moment.

The other 14 consumers have merged since. The counts have inverted:

| | canonical | mismatches |
|---|---|---|
| blockrun **main** | 270-line | **14** |
| blockrun **#469** | 360-line | **1** (this repo) |

So #469 is one repo from green and main is fourteen from it. Landing this makes #469 green; landing #469 makes main green.

## Packaging

`package.json` is untouched. The `!scripts/sync-brand-numbers.mjs` exclusion added by #306 and kept by #310 stays — that is what keeps the 360-line version's `child_process` import out of the published tarball. Verified that ClawRouter is the **only** consumer shipping `scripts/`: @blockrun/llm, @blockrun/nano-client and @blockrun/mcp publish `dist` only, so the import cannot reach their installs.

## Summary by CodeRabbit

* **New Features**
  * Added support for opting specific markers within fenced code blocks into rewriting and key validation.
  * Added reporting for skipped, non-live fenced markers without stopping processing.

* **Bug Fixes**
  * Improved handling of unterminated code fences.
  * Prevented replacement offsets from drifting during processing.
  * Limited rewriting to Git-tracked files in repositories.

### DIFF
--- a/scripts/sync-brand-numbers.mjs
+++ b/scripts/sync-brand-numbers.mjs
@@ -19,6 +19,7 @@
  * and wrap the WHOLE token, so a badge URL, its alt text and the prose number
  * can all regenerate from one key.
  */
+import { execFileSync } from "node:child_process";
 import { existsSync, lstatSync, readFileSync, writeFileSync, readdirSync } from "node:fs";
 import { join, relative, extname } from "node:path";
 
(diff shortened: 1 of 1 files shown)


VickyXAI:
Superseded by #311, which landed the same canonical script at 17:16 and also fixed the stale number the new skipped-marker warning surfaced. ClawRouter main is now `e5ed2807`, matching blockrun#469's canonical and the other 14 consumers. Closing rather than merging — this would be a no-op.

Issue #317, opened 2026-09-02
Pre-auth catch spans the payment-carrying send, so a lost response signs a second payment (payment-preauth.ts:131)


aurumflux20:
## Summary

In `src/payment-preauth.ts` the `try` at line 116 wraps not only the payment *signing* but the network send at line 123 that already carries a signed payment. Its `catch` at line 131 is commented *"Pre-auth signing failed — invalidate and fall through"*, which reads as though only `createPaymentPayload` / `encodePaymentSignatureHeader` can land there. `await baseFetch(preAuthRequest)` is inside the same block, so a rejected **send** lands there too — and the fall-through path signs and sends a **second, distinct payment**.

## The path

```ts
if (preAuthCovers) {
  try {
    const payload = await client.createPaymentPayload(cached.paymentRequired);   // 117
    const headers = httpClient.encodePaymentSignatureHeader(payload);            // 118
    …
    const response = await baseFetch(preAuthRequest);   // 123  ← send, payment attached
    if (response.status !== 402) return response;
    cache.delete(cacheKey);
  } catch (err) {                                       // 131
    cache.delete(cacheKey);
    if (err instanceof SpendPolicyError) throw err;
    // Pre-auth signing failed — invalidate and fall through.
  }
}
// clean, un-paid request → 402 → sign a NEW payload → send
const response = await baseFetch(request);              // 145
…
const payload = await client.createPaymentPayload(paymentRequired);   // 178  ← second authorization
return baseFetch(clonedRequest);                                       // 183
```

If `baseFetch` at 123 rejects **after** the gateway received and settled that payment — a connection reset, a socket hang-up, a TLS error, an upstream that closes after settling — the client cannot tell that from "the request never arrived." It takes the second reading, falls through, and pays again. One user request, two USDC charges.

## Why the existing guards don't cover this one

I checked both before writing this up, because you clearly already thought about double-charging:

- **The 30s response dedup** (`proxy.ts` header, line 16-17) keys on the request body and catches *OpenClaw retrying after a timeout* — a second proxy-level request. This double payment happens **inside one `payFetch` invocation**, so there is never a second proxy request and the dedup cache is never consulted.
[...]

PR #318, opened 2026-09-02
fix(desktop): stage ClawRouter 0.12.266 for release

### LINKED ISSUE #316: Desktop runtime pin goes stale on every release, silently falling back to a live npm install
Found while cutting v0.12.266.

`CLAWROUTER_PACKAGE_VERSION` (`apps/desktop/electron/core/runtime.ts`) is asserted equal to the root `package.json` version, so a release bump moves it. `findPinnedPackage()` then looks for that exact version inside the staged runtime — but the staged runtime comes from `apps/desktop/runtime/pnpm-lock.yaml`, which reaches `@blockrun/clawrouter` transitively through `@blockrun/clawrouter-codex`, and `stage:runtime` runs `pnpm install --frozen-lockfile`.

So on every release the constant is one version ahead of what the lockfile can stage. `findPinnedPackage()` misses, and `ensureNpmPackage()` falls through to `npm install @blockrun/clawrouter@<version>` at Connect time. It still works — the install is version-pinned, runs `--ignore-scripts`, and the resulting version is verified — but it needs the network and it is not the frozen runtime the pin exists to provide.

The ordering is genuinely circular: the lockfile cannot resolve a version that is not yet on npm, and the version is published from the release commit. So relocking is a post-publish step, and there is currently nothing that makes it happen or notices when it did not.

`tests/manager.test.ts` demonstrated the fallback rather than describing it — its fixture hardcoded `0.12.265`, and the v0.12.266 bump dropped it into the npm-install path with a misleading error. Fixed by importing the constant, but that only fixes the test.

Worth considering:

- A post-publish step that relocks `apps/desktop/runtime` and commits it, so a packaged build always stages the version it asks for.
- Or depend on `@blockrun/clawrouter` directly in `runtime/package.json` so the pin is explicit rather than a transitive accident.
- Either way, have `ensureNpmPackage()` say out loud when it falls back, so a packaged build that lost its frozen runtime is visible instead of merely slower.

`apps/desktop/tests/runtime-version.test.ts` currently asserts only that the lockfile stages exactly one copy — the equality assertion is deliberately absent, with the reason in a comment pointing here.

https://claude.ai/code/session_017wMqBG57KZjH7TKybHN8BS

KillerQueen-Z:
## Summary

- pin the frozen Desktop runtime to the now-published `@blockrun/clawrouter@0.12.266`
- make ClawRouter an explicit runtime dependency instead of relying on the Codex bridge transitively
- make the runtime-version test fail when the packaged Core differs from the root release
- add a reproducible `npm run dist:release` macOS ZIP command

Closes #316.

## Validation

- `cd apps/desktop && npm test` — 46 passed, 1 skipped
- `cd apps/desktop && npm run typecheck`
- `cd apps/desktop && npm run dist:release`
- ZIP integrity verified
- packaged runtime verified as `@blockrun/clawrouter@0.12.266`
- packaged DSH, Codex bridge, and Pi executables smoke-tested

The generated ZIP is unsigned because no Apple Developer ID identity is installed on the build host; it should be treated as the ARM64 preview artifact until signing/notarization is configured.

## Summary by CodeRabbit

* **Bug Fixes**
  * Desktop packaging now verifies that the staged runtime uses exactly the expected ClawRouter release, preventing mismatched or incomplete macOS builds.
  * Updated the desktop runtime to ClawRouter version 0.12.267.

* **Documentation**
  * Documented how to generate the macOS ZIP release artifact.

* **Tests**
  * Updated runtime lockfile checks to ensure only one ClawRouter version is staged.


VickyXAI (approved):
The structural fix here is the right one, and it is the option I would have picked: declaring `@blockrun/clawrouter` as a direct runtime dependency plus a pnpm override makes the pin explicit instead of an accident of whatever range `clawrouter-codex` happens to carry. `dist:release` is a good addition too.

I pushed `ea5808b` with two changes rather than sending it back.

**Relocked onto 0.12.267.** This pinned 0.12.266, which main moved past when the pre-auth double-charge fix shipped (#317).

**Moved the strict equality check to packaging time.** As written the two tests in `runtime-version.test.ts` cannot both hold on a release commit — the first ties `CLAWROUTER_PACKAGE_VERSION` to the root `package.json`, so it moves the moment the version bumps, and the second needs the lockfile to already resolve that version, which npm cannot serve until that same commit has published it. Every release would go red until a follow-up landed.

The invariant worth enforcing is a packaging one: a packaged build must stage the version the app asks for. That is now `npm run verify:runtime`, which `dist` and `dist:release` both run before electron-builder, with an error that names the two versions and the exact relock command. A stale runtime fails the build instead of silently falling back to a live `npm install` on first Connect. The unit test keeps the half that is always true — exactly one copy staged, which is what the override buys and what `findPinnedPackage()` actually depends on.

Net effect is what #316 asked for: the pin cannot go stale in a shipped build, and main never goes red for the release window.

Issue #319, opened 2026-09-03
Gateway-start plugin-id migration skips the bare `plugins.entries.clawrouter` that `plugins install` actually writes


twzrd-sol:
Since #313 the gateway-start plugin-id migration is gated on BlockRun-owned config
fields. That gate never matches the config an ordinary pre-rename install actually
has on disk, so those users are not migrated on the `npm update -g` + restart path.

### The two paths prove ownership differently

`clawrouter setup` proves it from disk (src/cli.ts:710-730):

```js
const legacyPackagePath = join(homedir(), ".openclaw", "extensions", "clawrouter", "package.json");
...
legacyBlockRunInstall = legacyPackage.name === "@blockrun/clawrouter";
```

The gateway-start path proves it only from config fields (src/index.ts:413-419):

```js
const legacyBlockRunInstall = Boolean(
  legacyEntry &&
    (["walletKey", "routing"]).some(
      (key) => key in legacyEntry || (legacyEntryConfig ? key in legacyEntryConfig : false),
    ),
);
```

### Why the config-field gate misses the common case

OpenClaw's installer commits `{ enabled: true }` for an `enabledByDefault` plugin and
nothing else. That is observable today: after `openclaw plugins install`, the entry is
exactly `"blockrun-clawrouter": { "enabled": true }`. A pre-rename install therefore
carries `"clawrouter": { "enabled": true }` with no `walletKey` and no `routing` --
the wallet lives at `~/.openclaw/blockrun/wallet.key`, not in `openclaw.json`, so the
fields that would prove ownership are absent unless the user hand-set them.

`prepareBlockRunPluginConfig` already handles a bare entry correctly once the flag is
true (src/openclaw-plugin-config.ts:61-63 copies a bare `enabled` boolean). The gate is
the only thing standing in front of it.

### Consequences on OpenClaw 2026.8.2

Verified end to end against a real `openclaw@2026.8.2` in an isolated HOME, on v0.12.267:

- Legacy `{ enabled: true }`: config ends as
  `{"clawrouter":{"enabled":true},"blockrun-clawrouter":{"enabled":true}}`. The stale key
  now explicitly enables OpenClaw's bundled router, a product the user never configured.
- Legacy `{ enabled: false }` (a pre-rename opt-out): config ends as
  `{"clawrouter":{"enabled":false},"blockrun-clawrouter":{"enabled":true}}`. The opt-out
[...]

Issue #305, opened 2026-09-02
fix(openclaw): avoid plugin ID collision with bundled @openclaw/clawrouter


KillerQueen-Z:
## Summary

OpenClaw ships a bundled plugin named `@openclaw/clawrouter` with plugin ID `clawrouter` (enabled by default since the OpenClaw 2026.7.1 release line). BlockRun's `@blockrun/clawrouter` declares the same plugin ID, so modern OpenClaw overrides the BlockRun plugin at runtime.

These are separate products: OpenClaw's provider uses `clawrouter/*`, `CLAWROUTER_API_KEY`, and `https://clawrouter.openclaw.ai`; BlockRun uses `blockrun/*`, a local proxy on `127.0.0.1:8402`, a self-custodial wallet, and x402 payments.

## Reproduction

Environment:

- macOS arm64
- Node.js 24.20.0
- OpenClaw 2026.8.2
- `@blockrun/clawrouter` installed under `~/.openclaw/extensions/clawrouter`

Steps:

1. Install/configure `@blockrun/clawrouter` with `clawrouter setup`.
2. Restart the OpenClaw gateway.
3. Run `openclaw plugins list --json`.
4. Check `http://127.0.0.1:8402/health` without separately starting the Desktop/standalone proxy.

Observed diagnostic:

```text
duplicate plugin id detected; global plugin will be overridden by bundled plugin
pluginId: clawrouter
source: ~/.openclaw/extensions/clawrouter/dist/index.js
```

The only loaded `clawrouter` plugin is:

```text
name: @openclaw/clawrouter
origin: bundled
status: loaded
providerIds: [clawrouter]
```

The local BlockRun proxy was not started by the gateway (`127.0.0.1:8402` was not listening).

## User impact

- Installation/setup can appear successful even though the BlockRun plugin is shadowed.
- Static `blockrun/*` provider configuration can still work if Desktop or a standalone CLI process starts port 8402 separately.
- Plugin-native BlockRun commands, tools, hooks, and gateway-managed proxy startup are not registered.
- Desktop status detection can report `Configured` by seeing the official `clawrouter` entry.
- Desktop disconnect/deactivation currently removes `clawrouter` from plugin entries/allowlists and can modify the official bundled plugin's configuration.
- There is no automatic wallet/payment or traffic handoff to OpenClaw's service; the provider namespaces and credentials remain separate.

## Suggested fix

[...]

PR #320, opened 2026-09-03
test(openclaw): version-pinned 2026.8.2 collision gate through the real installer

twzrd-sol:
Adds the OpenClaw 2026.7.1+ coexistence coverage asked for in #305, pinned to a version
that actually has the bundled `clawrouter` plugin.

### Why a new harness rather than the existing one

The unit tests cover the migration with a mocked `homedir`, so they cannot see the
installer -- and both ways this broke in production were installer-shaped: the id
collision itself (#305), and an install that aborts before the plugin is ever
registered. The existing container harness pins `openclaw@2026.5.4`
(test/docker-install/Dockerfile.openclaw-2026.5:3), which is the line *before* OpenClaw
started bundling its own `clawrouter`, so the collision does not exist there. Every other
harness floats on `openclaw@latest`, which cannot hold a version-specific regression
still.

This one pins `2026.8.2`, packs the tarball from the repo rather than a checked-in
filename, needs no Docker, and cleans up after itself.

### What it does

Inside a throwaway HOME (the operator's `~/.openclaw` is never touched):

1. seed a pre-rename `plugins.entries.clawrouter` entry
2. install the packed tarball through `openclaw plugins install`
3. boot the real gateway
4. assert both routers coexist

Scenario A covers a legacy entry left enabled; Scenario B covers a pre-rename opt-out.

    npm run test:e2e:openclaw-collision

`OPENCLAW_E2E_VERSION` changes the pin, `OPENCLAW_E2E_OPENCLAW` points at an existing
`openclaw.mjs` to skip the install, `OPENCLAW_E2E_KEEP=1` preserves the sandbox.

### Result on v0.12.267 against real openclaw@2026.8.2

11 of 16 checks pass. The #307 contract holds end to end: the bundled router keeps id
`clawrouter`, BlockRun is `blockrun-clawrouter`, the `blockrun` provider registers
(not merely present in config), and the x402 proxy activates and answers `/health` with
a wallet.

The 5 red checks are the stale-legacy-key behaviour filed as #319, not gate noise. They
are left red deliberately rather than weakened to green: the evidence gate they land on
is a deliberate choice in #313, and narrowing or widening it is your call, not something
a test should paper over. If you would rather this merge green, say which way you want
#319 resolved and I will fold the fix into this PR or adjust the assertions to match the
intended semantics.

## Summary by CodeRabbit

* **Tests**
  * Added release-gate coverage for OpenClaw plugin collision handling and legacy configuration migration.
  * Added verification that the BlockRun proxy activates correctly after installation.
  * Added coverage confirming that previously disabled configurations remain disabled and continue preventing proxy startup.

* **Chores**
  * Added a command to build the project and run the OpenClaw collision end-to-end test.

### DIFF
--- a/package.json
+++ b/package.json
@@ -58,6 +58,7 @@
     "test:resilience:full": "npm run test:resilience:errors && npm run test:resilience:lifecycle && npm run test:resilience:stability:full",
     "test:e2e": "tsx test/test-e2e.ts",
     "test:e2e:tool-ids": "npm run build && tsx test/e2e-tool-id-sanitization.ts",
+    "test:e2e:openclaw-collision": "npm run build && tsx test/openclaw-plugin-collision-e2e.ts",
     "test:docker:install": "bash test/run-docker-test.sh",
     "test:docker:edge-cases": "bash test/docker/run-edge-cases.sh",
     "test:docker:integration": "docker compose -f test/integration/docker-compose.yml up --build --abort-on-container-exit --exit-code-from integration"
--- a/test/openclaw-plugin-collision-e2e.ts
+++ b/test/openclaw-plugin-collision-e2e.ts
@@ -1,0 +1,397 @@
+/**
+ * Release gate: OpenClaw 2026.8.2 plugin collision/migration path (#305, #307).
+ *
+ * Exercises the REAL installation path end to end — no mocks:
+ *   1. isolated temporary HOME (the operator's ~/.openclaw is never touched)
+ *   2. seeds a legacy `plugins.entries.clawrouter` entry (pre-rename BlockRun)
+ *   3. installs the npm tarball packed from this repo via `openclaw plugins install`
+ *   4. runs `openclaw gateway` and asserts BOTH routers coexist:
+ *        - OpenClaw's bundled router keeps plugin id `clawrouter` (enabled, provider loaded)
+ *        - BlockRun is plugin id `blockrun-clawrouter`
+ *        - the legacy entry migrates away without clobbering either router
+ *        - BlockRun's x402 proxy ACTIVATES (GET /health answers with a wallet)
+ *   5. second scenario: a pre-rename opt-out (`enabled: false`) survives the
+ *      installer's {enabled:true} default, the bundled router stays enabled,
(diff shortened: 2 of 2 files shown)


twzrd-sol:
Prettier failure fixed in `a36bfab` — the new file was not formatted, so
`Lint & Typecheck` was red. Formatting only.

On the other bot finding, "make the default gate pass": leaving it as-is is
deliberate, and I would rather you decide than have me pick. The five red
checks are the stale-legacy-key behaviour in #319, not a defect in the test.
Weakening the assertions to green would encode the current behaviour as
intended, and the whole point of the gate is that it currently is not.

Three ways to land it, your call:

1. Merge as-is and let it stay red until #319 is resolved. Honest, but a red
   gate in CI is noise you may not want.
2. I fold the #319 fix into this PR, and the gate merges green.
3. I quarantine it behind an env flag or a separate script so it does not run
   in the default job, and it goes green once #319 lands.

I lean toward 2 if you agree with the direction in #319, since the fix is
small: have the gateway-start path use the same on-disk ownership proof that
`clawrouter setup` already uses.

VickyXAI:
Closing this one, and thank you for it — #319 came out of running it, and that fix shipped in v0.12.274 (1d9d51f) along with #329.

On the harness itself: it hard-owns port 8402 and refuses to run when something else holds it, with a comment explaining why (`BLOCKRUN_PROXY_PORT` does not survive the gateway). That makes it unrunnable on any machine where ClawRouter is actually running, which is every maintainer's — I could not run it here without stopping the operator's proxy. It is also not wired into CI (`ci.yml` runs build/typecheck/test only), and the 2026.8.2 pin is already behind 2026.9.2. A release gate that neither CI nor a developer can run is not a gate.

The regression it was built to catch is now covered in the default suite by `src/index.plugin-id-migration.test.ts` — both shapes you reported, including the inverted opt-out, verified to fail without the fix and to run in under two seconds.

If you want to bring this back: make the proxy port injectable rather than compiled in so the harness can take a free one, and wire it into a CI job. Then it is a gate rather than a script.

PR #322, opened 2026-09-03
fix(spend-control): one process-wide ledger shared by proxy, Polymarket, doctor and /policy

twzrd-sol:
#304 disclosed that the proxy (src/proxy.ts), the Polymarket tool
(src/polymarket/spend-policy.ts) and doctor each construct their own `SpendControl`.
Amount windows are therefore enforced per surface, and `saveHistory()` is
last-writer-wins on disk: LLM spend through the proxy does not count against a
Polymarket order, and either can overwrite the other's history.

Your point 2 on #303 asked for the primitive that fixes this -- have `startProxy`
expose the live `SpendControl`. #303 answered it locally, with a `liveSpendControl`
module variable in `index.ts` that only `/policy` could see. This replaces that local
answer with the general one: `getSharedSpendControl()` in `spend-control.ts` is the
single default, and `startProxy`, `buildPolymarketTool`, `createDoctorX402Client` and
`createPolicyCommand`'s `liveControl` all resolve to it when nothing is injected. The
module variable is deleted, not kept alongside.

Injection is unchanged. Every surface still accepts an explicit `spendControl`/`deps`;
only the default moved. Tests swap the instance with `setSharedSpendControl()` and
restore the previous one in a file-level `afterEach`, so no test file can leak state
into another.

### Restart semantics, stated rather than implied

An in-process proxy restart keeps the singleton's history and windows but **re-reads
limits from disk**. A hand-edit to `spending.json` followed by a restart still applies,
exactly as it does today, while spend already recorded in the window keeps counting.
That is what `SpendControl.reloadLimits()` is for: limits only, history untouched, and
it refuses a malformed file exactly like the constructor does, then recovers once the
file is repaired. A write after a reload uses the reloaded state as its compare-and-swap
baseline, so the stale-writer refusal from #303 still holds across a restart.

### Cross-process is still a known limit

A CLI write while the gateway is running is still a whole-object replace from another
process. #303's widened landed-check reports that race; this PR does not solve it and
does not claim to. It is not a file watcher and not a cross-process lock.

### Coverage

- both directions across surfaces: LLM spend through the proxy instance refuses a
  Polymarket order over the cap, and a Polymarket order recorded on the singleton
  refuses a proxy payment over the cap
- `/policy` reaches a surface it never touched before: blocking a payee through
  `runPolicyCommand` is enforced by a Polymarket order placed with no deps at all
- restart: edit the store, restart, assert the new limit is enforced while the
  pre-restart history record still counts
- `reloadLimits` fail-closed on a malformed file, and recovery after repair

No new dependencies. `tsc`, `eslint` and `vitest` clean: 941 tests pass, 1 skipped.

## Summary by CodeRabbit

- **New Features**
  - Unified spending controls across proxy payments, Polymarket actions, and doctor operations.
  - Spending history and policy enforcement now persist across in-process restarts.
  - Added safe reloading of spending limits from storage without losing active history.
  - Policy limits remain active when policy files cannot be read.

- **Bug Fixes**
  - Prevented separate spending ledgers from bypassing shared limits.
  - Ensured policy updates are consistently enforced across payment surfaces.
  - Improved handling of malformed, truncated, or unreadable spending policy files.

### DIFF
--- a/src/doctor.ts
+++ b/src/doctor.ts
@@ -31,7 +31,7 @@ import {
 import { getSolanaAddress } from "./wallet.js";
 import { getStats } from "./stats.js";
 import { getProxyPort } from "./proxy.js";
-import { registerSpendPolicyHook, SpendControl } from "./spend-control.js";
+import { getSharedSpendControl, registerSpendPolicyHook, SpendControl } from "./spend-control.js";
 import { VERSION } from "./version.js";
 
 // Types
@@ -517,7 +517,7 @@ export function createDoctorX402Client(opts: {
   const publicClient = createPublicClient({ chain: base, transport: http() });
   const evmSigner = toClientEvmSigner(account, publicClient);
   const x402 = new x402Client();
-  registerSpendPolicyHook(x402, opts.spendControl ?? new SpendControl());
+  registerSpendPolicyHook(x402, opts.spendControl ?? getSharedSpendControl());
   registerExactEvmScheme(x402, { signer: evmSigner });
   return x402;
 }
--- a/src/index.lifecycle.test.ts
+++ b/src/index.lifecycle.test.ts
@@ -49,12 +49,14 @@ describe("plugin lifecycle", () => {
       __clawrouterStartupGeneration?: number;
       __clawrouterStartedWithEmptyConfig?: boolean;
       __clawrouterStartupPhase?: "idle" | "probing" | "starting" | "running";
+      __clawrouterSharedSpendControl?: unknown;
     };
     proc.__clawrouterProxyStarted = undefined;
     proc.__clawrouterDeferredStartTimer = undefined;
     proc.__clawrouterStartupGeneration = undefined;
     proc.__clawrouterStartedWithEmptyConfig = undefined;
     proc.__clawrouterStartupPhase = undefined;
+    proc.__clawrouterSharedSpendControl = undefined;
   });
 
   it("clears deferred proxy startup state during deactivate", async () => {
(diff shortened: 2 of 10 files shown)


twzrd-sol:
Both bot findings were real. Fixed in `8a2fed5` and `11dbe7b`.

**Torn `spending.json` widened the policy on reload.** `load()` throws
`MalformedSpendPolicyError` for a bad policy list, which the reload already
refused on, but a read or parse failure took the other path: log and return
`null`, indistinguishable from "no file yet". The reload then cleared the
refusal state and set `limits = {}`, so a truncated file plus one in-process
restart left the proxy with no caps and no lists. The constructor can start
fresh on an unreadable file because it has nothing to lose; a reload cannot.
The distinction now lives at the storage boundary as
`UnreadableSpendPolicyError`, so `null` means only "nothing stored". Startup
behaviour is unchanged; the reload keeps the last known-good policy in force.
Test added red first.

Refusing outright on a torn file would be stricter, and I left it out on
purpose: that changes startup semantics beyond this PR. Say the word if you
want it and I will send it separately.

**The ledger was module-local while claiming to be process-wide.** Fair catch.
A global install and an npm-projects install can both resolve in one gateway,
and two module copies would each hold a ledger. It now lives on `process` as
`__clawrouterSharedSpendControl`, matching `__clawrouterProxyStarted` and the
other startup flags in index.ts. The test pins where the instance lives rather
than simulating a dual load, since loading two isolated module copies under
vitest tests the loader more than the ledger.

943 tests, typecheck, lint and prettier clean.

twzrd-sol:
Follow-up on the fail-open, because my own evidence was weaker than I claimed.

The tests I added for it used an in-memory stub that throws
`UnreadableSpendPolicyError` on demand. That proves the branch is handled, not
that a truncated file on disk reaches it — and "the file is not what the parser
expected" is exactly what a stub cannot reproduce. In my comment above I said a
torn file surfaces to the operator rather than silently no-op'ing; that was read
off the code path, not executed.

So I executed it, against the real `FileSpendControlStorage` with a real
half-written file (`16288ff`). Three of four go red against the pre-fix code, and
one is worse than the bug this PR set out to fix:

**`clawrouter policy limit daily 2` against a torn spending.json used to report
success and overwrite the file.** `result.isError` was `undefined`. `load()`
returned null, `saveLimits` saw `expect` match a null read, and the write landed,
discarding whatever payee or asset lists sat in the unreadable part. The person
most likely to hit that is an operator repairing a corrupted policy file. It now
reports the failure and leaves the file byte-identical.

The history save had the same shape: a torn file was previously rewritten from
limits that had just failed to load. Also now left alone.

The fourth test pins startup as deliberately unchanged — a torn file must not
take the constructor down, and still begins with no limits.

947 tests, typecheck, lint and prettier clean. Worth noting `tsc` caught a type
error in the new tests that all 947 passed straight through, which is a good
argument for the typecheck gate covering `*.test.ts`.

twzrd-sol:
Closed the remaining duplicate-module `/policy` gap from the outside-diff review in `7a9cb3f`. `liveControl` now keys off the process-wide proxy-started state instead of this module copy’s local handle, so a second plugin module reaches the same live ledger. Added a registration-level regression that sets a process-wide running proxy with no local handle and proves `/policy` reads the shared daily limit. Validation: 122 focused tests passed; typecheck, ESLint, Prettier, and `git diff --check` pass.

VickyXAI:
Superseded by #331 — your commits, rebased onto `main` (they had gone stale against the v0.12.268 API-key rail), with the review fixes on top. Your authorship is preserved on every original commit.

Opened as a new PR rather than updated in place because this branch lives on your fork: a rebase rewrites history, and GitHub refuses a maintainer force-push here (`remote rejected ... permission denied`) even with "allow edits by maintainers" on. Nothing was pushed to your branch.

Thanks — the review notes are in the new PR description.

twzrd-sol:
Rebased onto current upstream main after #20459/#ab380. Current tip: `d405603`. The duplicate-module `/policy` fix and lifecycle regression remain intact. Local verification on rebased tree: 122 focused tests passed; CI lint/typecheck and build/test are still running; scanner, Desktop, and CodeRabbit are green.

twzrd-sol:
Superseded by #331 (VickyXAI rebase + review fixes, already merged). Closing this original.

PR #325, opened 2026-09-03
fix(proxy): invoke onPayment after confirmed settlement

### LINKED ISSUE #321: `ProxyOptions.onPayment` is declared and documented but never invoked
`ProxyOptions.onPayment` is declared and documented, but nothing ever calls it, so a
caller that supplies it silently never hears about a settlement.

Declared at src/proxy.ts:1497:

```ts
onPayment?: (info: { model: string; amount: string; network: string }) => void;
```

Documented as a working example at docs/configuration.md:353:

```js
onPayment: (info) => console.log(`Paid ${info.amount} for ${info.model}`),
```

But `git grep -n onPayment -- src` returns only that declaration. The one other hit in
`src/` is `x402.onPaymentCreationFailure` at src/spend-control.ts:966, a different
symbol. There is no call site anywhere in `src/`, so the callback cannot fire on any
path -- it is passed in test/test-e2e.ts:840 and never invoked there either.

Matches observation: it did not fire on real settlements (three paid completions plus an
MCP call) while other spend accounting updated normally.

Either wiring it where a payment is confirmed, or dropping it from the type and the docs,
would be an improvement over a documented option that cannot work. Happy to send whichever
you prefer.

twzrd-sol:
Closes #321.

`ProxyOptions.onPayment` is public and documented, but the custom pre-auth fetch path never invoked it. Put the notification at the layer that owns both paid paths:

- normal 402 -> signed retry
- cached pre-auth -> paid first request

It fires only when the paid response carries the v2 `PAYMENT-RESPONSE` or v1 `X-PAYMENT-RESPONSE` settlement header. A rejected 402 does not fire it. Observer exceptions are contained after settlement so they cannot turn a paid response into a retry and risk a second charge.

Coverage proves normal and cached paths notify exactly once, a rejected cached payment does not add a notification, and a throwing observer does not trigger a retry.

Validation: 935 tests passed, 1 skipped; typecheck, ESLint, Prettier, and diff check pass.

## Summary by CodeRabbit

* **New Features**
  * Added payment notifications for settled payments, including the payment model, amount, and network.
  * Payment notifications now work for both pre-authorized payments and standard payment retries.
  * Proxy payment observers are now notified when payments are completed.

* **Bug Fixes**
  * Observer errors no longer cause successful payments to be retried.
  * Prevented duplicate notifications when cached pre-authorizations are rejected and refreshed.

### DIFF
--- a/src/payment-preauth.test.ts
+++ b/src/payment-preauth.test.ts
@@ -84,7 +84,10 @@ function fakeGateway() {
         ctl.rejectNextPaid = false;
         return challenge402(); // underpayment rejected
       }
-      return new Response(JSON.stringify({ ok: true }), { status: 200 });
+      return new Response(JSON.stringify({ ok: true }), {
+        status: 200,
+        headers: { "payment-response": "settled" },
+      });
     }
     return challenge402();
   });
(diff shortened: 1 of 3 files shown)


twzrd-sol:
Withdrawing this on our side — #321 stays open, and the call on whether you want
a fix at all is yours rather than ours to pre-empt.

Two reasons, and the second is worth keeping even if the first is just taste.

**It is unrequested.** #321 asked a question and offered either a wiring or a
removal. Sending the code before you answered skipped that, which is not how we
want to show up on your repo.

**More importantly, I do not think it is proven.** The notification is gated on
the paid response carrying `payment-response` / `x-payment-response`:

```ts
const settled =
  response.headers.has("payment-response") || response.headers.has("x-payment-response");
```

Those header names appear nowhere else in `src/` — that line is their only
occurrence outside the tests. And `proxy.ts` has no `upstream.headers.get(...)`
or `.has(...)` anywhere, so nothing in ClawRouter has ever read a header off a
paid upstream response. The tests go green because `fakeGateway()` was changed
in this PR to attach `payment-response: settled` to its 200. That proves the
wiring, not that the gate ever opens against the real upstream.

If it does not, the callback stays silent — which is the exact symptom #321
reported, moved one layer down and now covered by passing tests. That is a worse
failure mode than the one it replaces, because it looks fixed.

The codebase already has a signal it trusts for this: the `onAfterPaymentCreation`
hook, which writes the actual amount into the request-scoped store that the
logging path reads after `payFetch` (`proxy.ts:27`) — the same hook you pointed
at on #205 as what the builder-code stamp and chain logging use. It fires on
creation rather than confirmed settlement, so it is not a drop-in for what this
PR was reaching for, but it is the mechanism with evidence behind it.

So the open question on #321 is really: is there a response-level settlement
signal that the BlockRun upstream actually emits? You would know that
immediately, and we would be guessing. Happy to send a version built on whatever
you say the real signal is.

VickyXAI:
Superseded by #330 — your commits, rebased onto `main` (they had gone stale against the v0.12.268 API-key rail), with the review fixes on top. Your authorship is preserved on every original commit.

Opened as a new PR rather than updated in place because this branch lives on your fork: a rebase rewrites history, and GitHub refuses a maintainer force-push here (`remote rejected ... permission denied`) even with "allow edits by maintainers" on. Nothing was pushed to your branch.

Thanks — the review notes are in the new PR description.

PR #338, opened 2026-09-04
feat: complete account API service forwarding and polling

KillerQueen-Z:
Complete API-key service support through the local ClawRouter proxy: account requests retain upstream status/body/task URLs across service routes, while specialized media routes support first-response async jobs and credential-free result downloads. Account 402/429 errors remain account errors and never invoke x402 or switch to a wallet.

Address the review's account-cache issue by overriding Cache-Control to no-store on authenticated passthrough responses. Regression coverage switches account keys on the same proxy port/URL and checks both fresh ownership data and the cache policy. Empty, malformed or unreadable configured key files now stop startup instead of silently selecting another key or an old funded wallet; explicit account mode does not read wallet fields.

Accepted jobs recover from temporary gateway poll failures by GETting the same signed URL, never resubmitting paid creation requests. Correct timeout billing text, bundled skills and README registration/Keys/Credits/Activity instructions. Document source-versus-release service coverage, independent balances and logout/wallet-switch behavior. Regenerate the checked-in distribution.

Validation:
- Full suite: 1,034 passed, 1 skipped. Typecheck, ESLint, build, dist smoke and offline brand checks pass. One unrelated 5s auth-fixture timeout disappears with bounded workers; test assertions/timeouts were not relaxed.
- 73 focused account/passthrough regressions pass, including the review's no-store case and temporary poll recovery.
- Clean npm installation, CLI help and the installed proxy pass; the combined five-product HTTP run against merged Enterprise gateway source has 72 requests with simulated meter/debit reconciliation.

Local providers/storage are fixtures. No claim of every live model/service or real wallet settlement, and no package publication or deployment.

## Summary by CodeRabbit

- **New Features**
  - Added support for BlockRun account API keys with prepaid-credit billing alongside wallet-based x402 USDC payments.
  - Account mode now supports chat, media, search, RPC, partner, and data services where available.
  - Added automatic polling for asynchronous media and music jobs, including recovery from temporary gateway errors.
  - Account-authenticated requests preserve service responses and support standard 401, 402, and 429 error handling.

- **Documentation**
  - Updated setup, billing, authentication, wallet switching, pricing, and service guidance across the README and skills documentation.
  - Clarified account-key handling, polling behavior, and credit management.


VickyXAI:
Landed on `main` as bef780f, rebased onto current main with your authorship preserved (`Co-authored-by: Fsocietyhhh`). Thank you — the security half of this is work we did not have and should have.

Kept in full: the origin pinning and `redirect: "error"` (a redirect is the quiet way a bearer credential leaves the host it was minted for), `normalizeApiKeyBase`'s HTTPS/no-credentials check, stripping payment headers by pattern rather than exact name, `no-store` on authenticated responses, the async 202+poll_url completion that polls the signed URL instead of resubmitting a paid job, and making a malformed key fatal instead of skipped.

Four things changed on the way in, all of them consequences of `main` having moved since you cut the branch:

1. **Kept the usage-journal write in the passthrough.** Your early return said account billing is authoritative in the portal. That was true when you wrote it and isn't now — `x-blockrun-cost-usd` shipped today and 934bea4 consumes it, so the charge is knowable locally. Returning early would have re-hidden Surf/Exa/prediction-market spend from `/stats` on the API-key rail, which is the bug v0.12.271 had just fixed. Verified after merging: surf logs $0.0075 and exa $0.01, matching what the gateway actually charged.

2. **`no-store` is keyed on the credential, not the route.** You set it on your passthrough only, but the partner prefixes (`/v1/phone`, `/v1/surf`, `/v1/pm`) are matched earlier and are equally account-authenticated — under the original placement they kept upstream's `public, max-age=3600`. Your own regression test caught this: it asserts `no-store` on `/v1/phone/numbers/owned`, which the partner route serves. Good test.

3. **Empty `BLOCKRUN_API_KEY` is "unset", not "malformed".** Keying on `!== undefined` meant `FOO=""` — the ordinary way to clear a variable in CI and container images — hard-failed startup. Empty and whitespace-only now fall through; present-but-malformed stays fatal, which keeps your intent without breaking pipelines.

4. **One hole your change exposed elsewhere.** `index.ts` caught `resolveApiKey()`'s new throw with `.catch(() => undefined)` and fell through to `resolveOrGenerateWalletKey()`. Each change is defensible alone; composed, they cancel — the refusal that is the entire point of this PR would have minted a wallet instead. Fixed, along with the same warn-and-continue hole in `pluginConfig.apiKey`.

The README and `skills/*` changes were dropped as superseded: those files were rewritten earlier today (5375083 and others) for Solana-first ordering, corrected Surf pricing, and the account-credit surfacing, so re-applying yours would have reverted newer content. Nothing in them was wrong, just overtaken.

1050 tests, typecheck, lint and the DTS build pass; verified live against `api.blockrun.ai` in API-key mode with wallet mode unchanged. This will go out in the next release.

PR #343, opened 2026-09-05
fix: bind proxy reuse to credentials and restore account service parity

KillerQueen-Z:
Changing an API key or wallet could attach a new caller to an existing proxy billing the previous identity. The listen-race path also lost authentication-mode metadata. Reuse now checks the full-key fingerprint, gateway, wallet identity and chain in both paths; unverifiable old API proxies require a restart. API mode does not parse a leftover wallet.

Current main also lacked parts of the account service support covered by #338: native API paths, incremental Responses/Messages SSE, music/image-edit polling and account-specific cache controls. Restore those paths while preserving main's newer credit status, request journal and settled-charge header changes (through 5375083). Account fetches retain Request headers, remove payment proofs, reject foreign origins and redirects, and normalize account /v1 base URLs. Queued polling responses are fully consumed, addressing the review's response-body retention concern. Invalid configured keys fail instead of choosing another payer.

README/skills clarify registration, credit top-ups, restarting after credential changes, wallet-only settlement fees, actual Predexon payloads and Surf HTTP usage/pricing.

Validation: 1,061 local unit tests plus 5 lifecycle integration tests pass; independent lockfile install, build/dist smoke, typecheck, lint, formatting and brand-number checks pass. Added account/wallet/key/gateway/chain reuse tests including listen races; 43 service parity cases cover auth, errors, SSE, cache isolation and music polling. No live transfers or production activity/ledger reconciliation were performed in this pass.

This is a focused follow-up on current main to the review issues on #338, which remains open separately. Account passthrough retains main's x-blockrun-cost-usd charge accounting, verified by regression tests. Image generation, image editing, music and video also read the settled charge from the completed poll response instead of the initial submission or absent x402 hook. Dashboard activity remains authoritative where the gateway provides no settled cost.

## Summary by CodeRabbit

- **Documentation**
  - Clarified that paid features can use API account credit or an x402 wallet.
  - Documented payment-method selection, wallet/account changes, restart requirements, and payment troubleshooting.
  - Updated Surf instructions and pricing examples.
  - Clarified response payload handling for Predexon integrations.

- **Bug Fixes**
  - Invalid, blank, or unreadable API keys now report clear errors instead of silently falling back.
  - Improved API requests by preventing unintended redirects, cross-origin access, and payment-header reuse.
  - Added reliable handling for asynchronous jobs, cancellations, failures, and gateway errors.


VickyXAI:
Took the credential-binding half of this and landed it as 0554d1b — thank you, this was a real one and it was live.

I reproduced it on the shipped 0.12.272 before fixing: a proxy started with key A kept answering `/health` as key A, while a second process configured for key B printed the new key and reported `listening` on the same port. Every request went on billing A. Cross-account billing behind a success message, which is about the worst failure shape available here.

What landed: reuse now requires the masked label `/health` publishes to match, and a proxy that reports no label is refused rather than assumed to match — an unverifiable credential on a money path is not a match. Same-key reuse is untouched. Four tests, verified to fail against the previous code by reverting `proxy.ts` and re-running, since a guard that passes both ways guards nothing. One of them re-asserts the original wallet-vs-key refusal, so the new comparison joins it rather than replacing it. You are credited as co-author on the commit.

**The rest of this PR now conflicts and needs a rebase.** `main` moved a lot while it was open:

- **#338 landed** as bef780f, so the account-service parity work here is largely already in — native `/v1` passthrough, origin pinning, redirect refusal, `no-store`, async 202 polling, and fail-closed key resolution. Re-applying it would revert newer work rather than restore it.
- **v0.12.271 and v0.12.272 shipped** on top: the request-id journal, the settled-charge header, the 152x cost fix, the media $0 fix, `clawrouter reconcile`, and the credit warning.

Three things to know before rebasing, because they are places where this PR and current `main` disagree on purpose:

1. **The account passthrough keeps its `logUsage` write.** The early return in #338 said account billing is authoritative in the portal. That was true when written; `x-blockrun-cost-usd` now exists and the journal consumes it, so dropping the row would re-hide Surf/Exa/prediction-market spend from `/stats` — the bug 0.12.271 fixed.
2. **`no-store` is keyed on the credential, not the route.** The partner prefixes are matched earlier and are equally account-authenticated.
3. **Empty `BLOCKRUN_API_KEY` is treated as unset, not fatal.** `FOO=""` is the ordinary way to clear a variable in CI; present-but-malformed stays fatal.

If you rebase onto `main` and drop what has already landed, I expect this reduces to a fairly small diff. Happy to review it promptly.

VickyXAI:
Superseded. main already carries the origin pin, redirect refusal, header stripping by pattern, fatal-on-malformed key resolution, `normalizeApiKeyBase`, `pollApiKeyJob` and the key-fingerprint reuse guard — landed as #338/bef997c and 0554d1b, and this branch conflicts with them.

The one finding here that main did NOT have is real and is now fixed: the EADDRINUSE reuse path checked the payment chain and nothing else, so an API-key caller racing onto a wallet proxy reused it and got a handle reporting `authMode: "api-key"` while every request spent USDC — and its truthiness test on the reported wallet dropped API-key proxies, which publish an empty one. Both reuse paths run one validator now.

Shipped in v0.12.274 (1d9d51f), credited to you in the CHANGELOG.

PR #352, opened 2026-09-06
Fix Solana payment signing under Node/undici

0xCheetah1:
## Summary

Fixes a Solana x402 payment-signing fragility in bundled ClawRouter builds by removing the manual `content-length` header emitted by `@solana/rpc-transport-http`.

This only affects users on the **Solana payment chain**. Base-chain users and API-key users are not affected by this specific issue.

## Root Cause

The bundled Solana RPC transport currently includes:

```js
"content-length": body.length.toString()
```

That header is unnecessary in Node fetch/undici; undici can compute the request length itself. Keeping the bundled manual header leaves the Solana RPC signing path sensitive to body/header-length divergence if the upstream transport shape changes or non-byte-length-safe content enters that path. When that class of mismatch happens, the failure mode is either a Node/undici `fetch failed` error or a malformed/truncated Solana RPC request.

This PR does not claim to cover every Solana `fetch failed`. In particular, v0.12.271 fixed a separate reproduced Solana payment-signing failure caused by an unreachable Solana RPC endpoint via `CLAWROUTER_SOLANA_RPC_URL`. Users seeing Solana `fetch failed` should still check RPC reachability/configuration separately.

## Fix

- Add a tsup/esbuild plugin that strips only the manual Solana RPC `content-length` header from `@solana/rpc-transport-http/dist/index.node.mjs` during bundling.
- Fail the build if that targeted strip matches nothing, so a dependency reformat or upstream shape change cannot silently ship the header again.
- Add a dist smoke check so future publish builds fail if the manual header appears in shipped bundles.
- Regenerate dist output.

## Reproduction

On an affected Solana-chain install:

1. Install/start ClawRouter through OpenClaw.
2. Switch payment chain to Solana:

```sh
npx @blockrun/clawrouter chain solana
```

3. Ensure the Solana wallet has USDC/SOL for payment signing.
4. Send a paid proxy request through the local ClawRouter proxy:

```sh
curl -sS http://127.0.0.1:8402/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "openai/gpt-4o-mini",
    "messages": [{"role": "user", "content": "Reply with OK"}],
    "max_tokens": 8
  }'
```

With the patched bundle, the same Solana-paid proxy request succeeds and includes the Solana payment receipt headers. If a Solana `fetch failed` remains after this change, check Solana RPC reachability and `CLAWROUTER_SOLANA_RPC_URL`; that is a separate failure mode.

## Verification

Tested on Node 22.23.1/22.23.2:

- `npm ci`
- `npm run build`
- `npm run format:check`
- `npm run typecheck`
- `npm run lint`
- `npm test`
- confirmed the bad header is absent from `dist/index.js`, `dist/cli.js`, and `dist/router/index.js`

## Summary by CodeRabbit

- **Bug Fixes**
  - Prevented an incompatible HTTP header from being included in bundled builds, improving reliability for Solana payment requests.
  - Reduced the risk of request failures in Node.js and undici environments caused by the unwanted header.
  - Improved build verification so detection remains reliable regardless of formatting.


VickyXAI (commented):
Reviewed. The change is safe and I verified it works, but there is one bug in the guard itself, and the stated root cause did not reproduce.

## P1 — the strip and the guard fail together, silently

`String.replace` returns the source unchanged when the pattern misses, so the plugin no-ops on any upstream reformat. The smoke check keys on nearly the same literal, so it misses the same cases. Both mitigations fail at once, the build goes green, and the header ships — the exact regression the guard exists to prevent.

Reproduced with the only change being single-quoted keys upstream:

```
strip changed anything?  false
guard would flag it?     false
```

Broadening the guard regex is not the fix — `"content-length": true` appears twice in that file's `DISALLOWED_HEADERS` / `FORBIDDEN_HEADERS` maps, so a looser grep false-positives.

The fix is one line in the plugin: assert it matched.

```js
const contents = source.replace(/\n\s+"content-length": body\.length\.toString\(\),/, "");
if (contents === source) {
  throw new Error(
    "strip-solana-rpc-content-length matched nothing in " + args.path +
    " — @solana/rpc-transport-http changed shape. Update the pattern; do not ship."
  );
}
return { contents, loader: "js" };
```

That turns a silent dependency-bump regression into a build failure naming the file.

## P2 — the root cause as written does not reproduce

The PR says Node/undici "can reject that manually supplied header". I could not reproduce that on any Node available: **18.19.0, 20.5.0, 22.14.0, 22.21.1, 22.22.0, 22.23.2, 24.18.0, 24.19.0** all accept a *correct* manual `content-length`.

undici rejects only when the value disagrees with the body:

```
manual correct       OK
manual WRONG(+10)    THREW: Request body length does not match content-length header
```

`body.length` is UTF-16 code units, so it can only ever *under*-state, never over-state — and under-stating did not throw, it produced a silent empty response. Either way it needs a **non-ASCII body**, and this path sends base58 addresses and base64 transactions, where `body.length === Buffer.byteLength(body)`.

So the header is latent here rather than firing. **Keep the change** — letting undici compute the length is strictly more correct, and the failure mode when it does bite is a silent truncated request, which is worse than a throw. Just please reword the root cause, because v0.12.271 shipped a *different*, reproduced cause for the same `fetch failed` symptom: an unreachable Solana RPC during payment signing, fixed with `CLAWROUTER_SOLANA_RPC_URL`. Someone chasing the next Solana `fetch failed` should not stop at this PR believing it is covered.

## Verified

Built the branch on Node 22.23.2: header gone from `dist/{index,cli,router/index}.js`, bundle sizes 7.9MB / 8.0MB (no tsup-alias regression), committed `dist` reproducible byte-for-byte from a fresh build, typecheck and lint clean, 1088 tests passing. Removing the plugin makes the smoke check fail with a clear message — the guard does work for the case it covers.

Happy to merge once P1 is in. P2 is a description change, not code.

VickyXAI:
Reviewed this closely — thanks for chasing it, the underlying defect you found is real. But I don't think the build-time strip is the right shape, and I've opened #358 with the upstream fix instead.

**What's right here, verified:**
- The committed `dist/` **reproduces exactly** from a clean rebuild of this branch (path-normalized, byte-for-byte on all three bundles). For a fork PR shipping build output, that's the thing I most wanted to check, and it passed.
- The new smoke assertion is **not vacuous** — the pre-patch bundle does match its regex.
- Removing the header is safe: undici sets `content-length` itself, correctly and in bytes, with no fallback to chunked encoding (checked on Node 22.23.2 and 24.19.0).
- The defect is genuine: `body` is a string, so `body.length` is UTF-16 code units, not UTF-8 bytes. A 1-emoji body (53 units / 55 bytes) truncates and desyncs the connection.

**Why I'm not merging it as-is:**

1. **The stated failure isn't reachable here, and doesn't reproduce on the Node you named.** Every Solana RPC body on both paths is ASCII by construction — `getBalance`/`getTokenAccountsByOwner` take base58 addresses; the signing path (`@x402/svm`) sends `getLatestBlockhash`, `sendTransaction`, `simulateTransaction` etc. with base58, base64 and numbers. No user or model string ever enters a Solana RPC body. Against real undici with the actual payloads: HTTP 200 on Node 22.14.0 (undici 6.21), **22.23.2 (6.28)** and 24.19.0 (7.29). undici does not reject the header. The body is honest about this ("sensitive … **if** … non-byte-length-safe content enters that path"), but the title and the Reproduction section read as a fix for an observed failure.

2. **Upstream already deleted the line** — present in 5.5.1/6.3.1/6.5.0/6.7.0, gone from 6.10.0 on; `8.2.0`'s `createHttpTransport` is byte-identical to 5.5.1's minus that header. #358 takes that instead of carrying a permanent patch.

3. **This PR blocks that upgrade.** With kit 8 installed, the plugin hard-fails the build:
   ```
   ✘ [ERROR] strip-solana-rpc-content-length matched nothing in .../index.node.mjs
     — @solana/rpc-transport-http changed shape. Update the pattern; do not ship.
   ```
   The fail-closed condition is "a replacement occurred", so it breaks exactly when the dependency becomes correct. If we ever do keep a patch, the condition should be "the emitted header object has no `content-length`".

4. **The smoke guard can silently false-pass.** `scripts/smoke-dist.mjs:53` hardcodes the local identifier `body`, and esbuild demonstrably renames locals *inside this very function* — upstream's `await fetch(url, requestInfo)` ships as `await fetch(url2, requestInfo)` in `dist/cli.js`. The day the graph collides on `body`, the guard passes while the header ships.

5. **Minor:** `dist/router/index.js` never contained the transport at either end, so listing it as verification overstates coverage. And the patched bytes exist in no npm tarball, no lockfile entry, and the sourcemap carries the patched text — a `patch-package` file or an `overrides` pin is a reviewable artifact; a regex in `tsup.config.ts` isn't.

**One thing you were right about that I initially doubted:** the signing path *does* reach this transport — `@x402/svm` calls `createSolanaRpc` internally. So the title is accurate about which code path, even though the failure isn't reachable on it.

#358 needs a live Solana-rail payment before it can merge (the tests mock `@x402/svm` wholesale, so green means little there). If that verification fails and we need a local stopgap, I'd rather take a narrow `overrides` pin than build machinery — happy to revisit this branch then.

VickyXAI:
Closing this — the defect you found is real and it's fixed, just from the other end.

`@solana/rpc-transport-http@5.5.1` really does set `"content-length": body.length.toString()` where `body` is a string, so it's a UTF-16 code-unit count rather than UTF-8 bytes. Good catch. Upstream had already deleted that line: it's present in `5.5.1`, `6.3.1`, `6.5.0`, `6.7.0` and gone from `6.10.0` on, and `8.2.0`'s `createHttpTransport` is byte-identical to `5.5.1`'s minus that header. #358 takes the upgrade, so the header is gone at the source with no patch to maintain — and undici then computes the length itself, correctly and in bytes.

Two things that decided it against the build-time strip:

**It would have blocked the upgrade.** With kit 8 installed, the plugin hard-fails the build — `strip-solana-rpc-content-length matched nothing in .../index.node.mjs — @solana/rpc-transport-http changed shape. Update the pattern; do not ship.` The fail-closed condition is "a replacement occurred", so it breaks precisely when the dependency becomes correct.

**The failure isn't reachable on our payloads.** Every Solana RPC body on both paths is ASCII by construction — base58 addresses, base64 transactions, `commitment` strings. Against real undici with the actual payloads it's HTTP 200 on Node 22.14.0 (undici 6.21), 22.23.2 (6.28) and 24.19.0 (7.29); undici doesn't reject the header. It only breaks when the count is wrong, which needs non-ASCII in the body.

For the record, the parts of this PR that were right: the committed `dist/` **reproduced exactly** from a clean rebuild of your branch, byte-for-byte on all three bundles after path normalization — that's the check that matters most on a fork PR shipping build output, and it passed. The smoke assertion wasn't vacuous either; the pre-patch bundle does match its regex. And removing the header is safe: undici sets it itself with no fallback to chunked encoding.

One thing worth carrying forward if a build-time patch ever is the right tool: `scripts/smoke-dist.mjs` hardcoded the local identifier `body`, and esbuild demonstrably renames locals inside that very function — upstream's `await fetch(url, requestInfo)` ships as `await fetch(url2, requestInfo)` in `dist/cli.js`. A guard keyed on an identifier passes exactly when it's needed most.

Thanks for chasing this one — it sent us to a real fix. #358 is open and green; it's gated on a live Solana payment check before it merges, since the tests mock `@x402/svm` wholesale.

PR #358, opened 2026-09-07
deps: @solana/kit 5.5.1 → 8.2.0 (takes the content-length fix from upstream)

VickyXAI:
Fixes the `content-length` defect #352 targets, by taking the fix upstream already shipped instead of patching it out of the bundle.

## The defect

`@solana/rpc-transport-http@5.5.1` (`dist/index.node.mjs:73,81`):

```js
const body = toJson ? toJson(payload) : JSON.stringify(payload);
...
"content-length": body.length.toString(),
```

`body` is a string, so `.length` is UTF-16 code units, not UTF-8 bytes. Measured against a local server: a 1-emoji body is 53 units / 55 bytes, and the request fails with `Unexpected end of JSON input`, then the connection desyncs. The library's own `DISALLOWED_HEADERS` and `FORBIDDEN_HEADERS` tables (lines 8, 17) list `content-length` as a header callers must not set — it then sets it anyway.

## Why upstream, not a build-time patch

Upstream deleted the line: present in `5.5.1`, `6.3.1`, `6.5.0`, `6.7.0`; gone from `6.10.0` on. `8.2.0`'s `createHttpTransport` is byte-identical to `5.5.1`'s minus that header. undici then computes `content-length` itself, correctly and in bytes — verified on Node 22.23.2 and 24.19.0, with no fallback to `Transfer-Encoding: chunked` (some RPC providers reject chunked).

**#352's build-time strip actively blocks this upgrade.** With kit 8 installed, its esbuild plugin hard-fails the build:

```
✘ [ERROR] strip-solana-rpc-content-length matched nothing in .../index.node.mjs
  — @solana/rpc-transport-http changed shape. Update the pattern; do not ship.
```

Its fail-closed condition is "a replacement occurred", so the build breaks precisely when the dependency becomes correct.

## Why this needs `overrides`

`@x402/svm@2.25.0` (latest) still pins `@solana-program/token ^0.9.0`, `token-2022 ^0.6.1`, `compute-budget ^0.11.0` — all peering on `@solana/kit ^5.0`. Their current releases (`0.16.1`, `0.16.1`, `0.18.1`) have moved to `^8.0.0`. So the clean bump is gated on the x402 SDK, not on this repo.

The overrides are **merged into** the existing block, not replacing it — the 13 entries already there are security pins (`ws`, `tar`, `postcss`, `basic-ftp`, `ip-address`, `brace-expansion`, `jayson/uuid`, …), all still enforced.

Forcing x402's transitive deps forward is safe because `@x402/svm` uses exactly four things from them — `findAssociatedTokenPda`, `getTransferCheckedInstruction`, `getSetComputeUnitLimitInstruction`, and the two program-address constants. Driven with identical inputs on both trees, every byte matches:

| | old (token 0.9.0 / cb 0.11.0) | new (token 0.16.1 / cb 0.18.1) |
|---|---|---|
| ATA (bump) | `FGETo8T8…ra8B` (254) | identical |
| transferChecked data | `0c87d612000000000006` | identical |
| transferChecked accounts | 4 accounts, roles `1,0,1,0` | identical |
| computeBudget data | `02400d0300` | identical |
| TOKEN / TOKEN_2022 program | unchanged | unchanged |

npm resolves **one deduped `@solana/kit@8.2.0`** — no second copy, which matters because `noExternal: [/.*/]` would bundle both.

## Verified

- `tsc --noEmit` clean
- 1088 passed / 1 skipped
- `npm run build` + dist smoke green
- built bundles carry **zero** manual `content-length` lines, with no patching of any kind

## NOT verified — needs a human before this ships

**A live Solana payment.** The only tests touching the signing path mock `@x402/svm` wholesale (`src/proxy.solana-rpc-override.test.ts:22,32`), and nothing in `src/` calls the token builders directly — they live inside x402's compiled code. Green tests say nothing about the real money path. One real Solana-rail paid request should gate the merge.

`dist/` is deliberately not regenerated here to keep the diff reviewable; `publish.yml` rebuilds from source anyway (`npm ci` → build → publish). Note main's committed `dist/` is separately stale.

## Summary by CodeRabbit

- **Chores**
  - Updated underlying Solana components to newer compatible versions.
  - No direct user-facing feature or behavior changes are included in this release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "@blockrun/clawrouter",
-  "version": "0.12.262",
+  "version": "0.12.276",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "@blockrun/clawrouter",
-      "version": "0.12.262",
+      "version": "0.12.276",
       "license": "MIT",
       "dependencies": {
         "@blockrun/llm": "^3.10.0",
@@ -16,7 +16,7 @@
         "@polymarket/clob-client-v2": "1.0.8",
         "@scure/bip32": "^2.0.1",
         "@scure/bip39": "^1.5.0",
-        "@solana/kit": "^5.0.0",
+        "@solana/kit": "^8.0.0",
         "@x402/core": "^2.9.0",
         "@x402/evm": "^2.9.0",
         "@x402/fetch": "^2.9.0",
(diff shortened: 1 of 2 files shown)


VickyXAI:
Reviewed with a built comparison of both trees, not just the diff. **The signing risk is not real; the bundle cost is. My recommendation is to hold this, and I've landed the guard that makes it safe to revisit.**

## The `^5.0.0` invariant does not apply here

It was written after 2026-03-06, where `@solana/kit` and `@x402/svm` resolved to different majors and left two runtime copies of `@solana/signers` / `@solana/transactions` disagreeing about signature state. That does not recur:

- In the PR-built bundle there is **exactly one** copy of `@solana/signers`, `@solana/transactions`, `@solana/transaction-messages` and `@solana/kit`.
- Driving `@x402/svm`'s real client path (`ExactSvmScheme.createPaymentPayload` and the V1 scheme) with a fixed keypair, fixed blockhash and fixed memo against a stubbed RPC produces **byte-identical signed wire transactions** under kit 5.5.1 and 8.2.0.
- All 32 symbols `@x402/svm` and ClawRouter import from kit / `@solana-program/*` exist in 8.2.0. Zero removed, zero renamed.
- Peers are clean. `@x402/svm@2.21.0` peers on `@solana/kit: ">=5.1.0"`, not `^5.0`, so kit 8 satisfies it unmodified. The three `@solana-program` overrides are **required**, not incidental — 0.9.0/0.6.1/0.11.0 each peer on kit `^5.0`. `npm ls` reports zero invalid and zero unmet non-optional peers.
- The overrides block really is a merge: all 13 security pins survive with identical values, pure 4-key append.

The duplicated packages (20 copies of `@solana/errors`, 9 of `codecs-core`, …) are all stateless — `@solana/errors` identifies by `e.name === "SolanaError"`, never `instanceof`; nothing anywhere does `instanceof SolanaError`.

## What it costs

`dist/cli.js` **8.00 MB → 10.11 MB (+26.4%, +2,211,641 bytes)**. Both refs built from `git archive` under the same toolchain; the `main` build reproduced the committed dist byte-for-byte, so the comparison is sound. That leaves 1.88 MiB under `MAX_BUNDLE_BYTES`, a canary written to catch "a whole extra copy of something" — after this it can barely see one. Two fixes were tried and both failed: pinning all 20 `@solana/*` subpackages via `overrides` changed nothing, and an esbuild `onResolve` canonicaliser made it *worse* (11.36 MB).

## What it buys

On this repo, today: nothing I can demonstrate. #352's own body describes the header as leaving the path "**sensitive to** … **if** … non-byte-length-safe content enters that path" — a hardening claim, not a reproduced failure; the Solana `fetch failed` that *was* reproduced was an unreachable RPC endpoint, fixed separately in v0.12.271. `body.length` under-states only for non-ASCII, and every Solana RPC body on this path is ASCII (base58 addresses, base64 transactions, ASCII method names), where `String.length` equals the byte length and undici accepts the header. Verified across Node 18/20/22/24: a correct manual `content-length` is accepted; only an over-stated one throws.

So this trades 26% of the bundle and most of the duplicate-detection margin for a defect that cannot currently fire.

## Also

`dist/` is committed here and this PR does not rebuild it, so the tracked bundle still inlines kit 5.5.1 — including the header the title says it removes. And the lockfile now conflicts with `main`, so every number in the PR body was measured against a lockfile that won't be the one that lands.

## Suggested path

Hold until `@x402/svm` ships a release whose own `@solana-program` deps are on `^0.16`/`^0.18` — that removes the override-driven nesting, and the bump becomes a one-line version change with no bundle cost. `@solana/kit@8.2.0` on its own resolves to a single `@solana/errors`; the duplication only appears when kit 8 meets `@x402/svm`'s older pins.

Meanwhile `guard/solana-single-copy` (`5b2f848`) rewrites the invariant as the property that actually held — `@solana/signers`, `@solana/transactions` and `@solana/transaction-messages` must each be inlined exactly once — and enforces it in `smoke-dist`, counting distinct install paths rather than marker lines, with a self-check so a rotted matcher can't report a clean bundle. It passes on `main`, passes on this PR's tree, and rejects a synthetic v5/v6 split. That is the thing that was missing: the old rule was a version pin nobody enforced.

**Remaining gap either way:** no live settlement was performed. The byte-identical result proves construction and signing agree between kit versions; it does not prove the facilitator accepts the transaction. One real paid Solana call through the built bundle before any release.

VickyXAI:
Converting to draft — blocked on upstream, not on anything in this branch. Tracked as #378, which records the exact unblock condition (`@x402/svm` shipping `@solana-program` deps on `^0.16`/`^0.18`), the two mitigations already tried and failed, and everything now settled so none of it gets re-derived: byte-identical signing across 5.5.1 and 8.2.0, all 32 imported symbols present, clean peers, and the 2026-03-06 failure mode proven absent.

That last one is now a build guard rather than a memory — #375 asserts single-copy `@solana/signers` / `@solana/transactions` / `@solana/transaction-messages` in `smoke-dist`, so when this does land, the thing the old `^5.0.0` pin was really protecting is checked mechanically.

Reopen for review when `@x402/svm` moves; at that point the `overrides` come out and this is a one-line bump with no bundle cost.

PR #357, opened 2026-09-07
feat(x402): opt-in TWZRD AutoGate after SpendControl (#355)

twzrd-sol:
Closes #355.

Opt-in TWZRD AutoGate on the existing x402 `onBeforePaymentCreation` hook. **Default off.** Not a re-open of withdrawn #218 (default-on vendor lock). SpendControl (#205 / #268 / #304) stays the vendor-neutral path; this only composes TWZRD when explicitly enabled.

## Behavior

1. Default: unchanged — no TWZRD AutoGate, no extra import.
2. `TWZRD_AUTO_GATE=1` (or `TWZRD_GATE_ENABLED=true`) calls `installTwzrdAutoGate` / `createTwzrdBeforePaymentHook` **after** `registerSpendPolicyHook` on the same client. SpendControl lists still run first.
3. Optional dependency `twzrd-x402-gate@0.9.3` so forks that omit it still install.
4. Fail open only when that package itself is `MODULE_NOT_FOUND`. Any other load/install error fails closed.
5. Attribution: `X-Twzrd-Caller: clawrouter/<version>` so a refuse is attributable.

## Docs

README + `docs/configuration.md`: opt-in, default off, how to enable.

## Tests

`src/twzrd-autogate.test.ts`: default path does not load or register; with the flag the hook is composed and invoked; missing package fails open; other errors fail closed.

Verified: `tsc --noEmit`, `eslint src/`, `vitest run` (1100 passed, 1 skipped), `npm run build` (dynamic `import("twzrd-x402-gate")` stays unbundled).

## Summary by CodeRabbit

* **New Features**
  * Added optional TWZRD AutoGate protection for EVM x402 payments.
  * Enable it with `TWZRD_AUTO_GATE=1` or `TWZRD_GATE_ENABLED=true`; it is disabled by default.
  * Payment checks run after SpendControl and can refuse payments before transaction signing.
  * Requests include a TWZRD caller identity stamp.
* **Documentation**
  * Added configuration reference, CLI help, and setup guidance for enabling AutoGate.
  * Documented behavior when AutoGate is unavailable or encounters loading errors.

PR #360, opened 2026-09-07
chore(deps): pin twzrd-x402-gate to 0.9.4

twzrd-sol:
Follow-up to #357. Opt-in stays off. SpendControl stays first. Not a re-open of #218.

Pins optional `twzrd-x402-gate` to **exactly `0.9.4`** (npm `latest`). `0.10.0` / `0.10.1` are deprecated on npm as unreproducible — do not use them. The lockfile entry is regenerated so it carries `0.9.4`'s published peer ranges (`@solana/kit >=5.1.0 <6`, `@solana-program/token >=0.9.0 <0.10`, `x402-solana >=2.1.0`) instead of the `0.9.3` ranges it had inherited; integrity unchanged (`sha512-Fgdoz…`), exports carry no `./unsafe`.

## What 0.9.4 actually does on our pre-sign hook

Verified against the installed package, not the docs. `createTwzrdBeforePaymentHook` defaults to a **wash-only engine**, which is not what the v0.12.277 notes described (those matched `0.9.3`'s full preflight):

- **Request path.** One `GET merchant_card/{payTo}` per payment, on **Solana and Base alike**. ClawRouter does not select the full preflight engine — no `POST /v1/intel/preflight`, no reputation score, no Solana/Base split. Only the `payTo` leaves the machine (not resource URL, price or chain).
- **Refusal is broader than a wash boolean.** `wash_flagged: true` aborts (`twzrd_wash_flagged`); a clean `payTo` **with full coverage** allows; `wash_flagged: false` with missing, partial or stale coverage also aborts (`twzrd_wash_unknown`). Unknown is not clean.
- **Attribution.** Every lookup carries `X-Twzrd-Caller: clawrouter/<version>@0.9.4` and `X-TWZRD-Integration: clawrouter/<version>` (plus `X-TWZRD-Run-Id`, `X-TWZRD-Client: twzrd-x402-gate/0.9.4`).
- **Outage promise, narrowed.** `TWZRD_FAIL_OPEN=false` governs only ClawRouter's outer 2 s timeout/error wrapper. The package itself converts a fast lookup failure — a quick `503`, `404`, `fetch failed`, invalid JSON, or its own 3 s `merchant_card` timeout — into allow **before our wrapper sees anything**, and ignores the `failOpen` option we pass. So `TWZRD_FAIL_OPEN=false` means "refuse on hang", not "refuse on outage". Fully enforcing refuse-on-outage needs a package fix (out of scope here). Docs now say so.

README, `docs/configuration.md`, the CHANGELOG `Unreleased` entry and the adapter's comments are corrected accordingly. No behaviour change in `src/twzrd-autogate.ts`; no version bump (release stays on the normal process).

## Verification matrix (real `twzrd-x402-gate@0.9.4` + current adapter, mocked HTTP)

| Case                                                           | Result                                                                          |
| -------------------------------------------------------------- | ------------------------------------------------------------------------------- |
| `wash_flagged=true` (Solana and Base)                          | Aborts `twzrd_wash_flagged`; signer calls: 0                                    |
| Clean, full coverage (Solana and Base)                         | Allows; one `GET merchant_card/{payTo}`, attributed; no preflight POST          |
| `wash_flagged=false`, missing / partial / stale coverage       | Aborts `twzrd_wash_unknown`; signer calls: 0                                    |
| Fast HTTP 503 / 404 / `fetch failed` / non-JSON, `TWZRD_FAIL_OPEN=false` | **Allows** (package-internal fail-open; our wrapper never warned)      |
| Hang past our budget, `TWZRD_FAIL_OPEN=false`                  | Aborts `twzrd gate did not answer within …ms`; signer calls: 0                  |
| Hang past our budget, default config                           | Allows with a warning                                                           |
| Package's own wash timeout shorter than our budget, `TWZRD_FAIL_OPEN=false` | **Allows** (same limitation from the other side)                    |

Tests load the real package through the adapter's own dynamic import and drive a real `x402Client` with a fake scheme as the signer; only `fetch` is stubbed. The suite skips cleanly if the optional package is absent (`npm ci` installs it in CI). `src/twzrd-autogate.test.ts`: 23 → 41.

## Checks

`npm ci` · `npm ls twzrd-x402-gate` → `0.9.4` · `npm test` → 1129 passed, 1 pre-existing skip · `npm run typecheck` · `npm run build` (+ dist smoke) · `npm run lint` · `prettier --check .` · `sync-brand-numbers --check` · `git diff --check` — all green.

Not touched: `dist/`. The committed bundle still carries the `@0.9.3` string in the "not installed" hint that `b4caf5f` changed in `src`; it picks up on the next rebuild.

## Summary by CodeRabbit

- **Documentation**
  - Updated TWZRD AutoGate guidance for version 0.9.4 and exact dependency pinning.
  - Clarified wash-only checks across Solana and Base, including flagged, unknown, partial, and stale coverage handling.
  - Documented payment lookup data and attribution headers sent with requests.
  - Clarified observe mode and fail-open behavior, including timeout and fast lookup failure handling.

- **Chores**
  - Pinned the optional TWZRD AutoGate dependency to version 0.9.4.

- **Tests**
  - Expanded coverage for payment checks, refusal conditions, network behavior, and timeout scenarios.


twzrd-sol:
Ready for review/merge when you are.

Checks are green; pin is exact `twzrd-x402-gate@0.9.4`. One qualification for docs/changelog readers: the AutoGate hook here is **wash-only** (`merchant_card/{payTo}` on Solana + Base) — it does not send resource/amount/chain into full reputation preflight. No overclaim.

Happy to address the CodeRabbit Base regression-test nit if you want it in this PR; otherwise fine as follow-up.

VickyXAI:
Closing, and TWZRD AutoGate is removed from ClawRouter in [v0.12.278](https://github.com/BlockRunAI/ClawRouter/commit/107c2d1) along with the `twzrd-x402-gate` optional dependency, the `TWZRD_*` flags and their docs.

Thank you for the verification in this PR — it is the reason for the decision. Verified against the installed package, `0.9.4` shows four things we will not carry inside `onBeforePaymentCreation`:

- **The engine changed under a patch bump.** `0.9.3` ran the full `/v1/intel/preflight` we documented; the same `createTwzrdBeforePaymentHook` call in `0.9.4` is a wash-only `GET merchant_card/{payTo}` against a different endpoint with a different refusal set, with no code change on our side. A dependency that redefines what it does in the pre-sign path on a patch release is not pinnable in a way that helps.
- **Our failure config was never honoured.** The package converts a fast `503` / `404` / `fetch failed` / non-JSON into allow internally and ignores the `failOpen` we pass, so `TWZRD_FAIL_OPEN=false` never meant refuse-on-outage. We documented a guarantee the package would not give.
- **Unknown is a refusal.** `wash_flagged: false` with missing, partial or stale coverage aborts. A recipient nobody has scored yet stops a user's payment.
- **Every lookup was attributed.** `X-Twzrd-Caller`, `X-TWZRD-Integration`, `X-TWZRD-Client` and a run id, sent from the code path that signs with the user's wallet.

ClawRouter's payment path stays vendor-neutral. `clawrouter policy` (SpendControl) is the pre-spend control we ship and support, and it is unaffected. This is a settled decision, not an invitation to iterate on the integration — please don't open further PRs adding it back under a different shape or flag.

Your unrelated contributions remain welcome and credited (#320, #303, #321/#325, #322/#331).

PR #367, opened 2026-09-08
desktop: visual refresh of the control plane

ramioca:
## Summary

Visual refresh of the ClawRouter Desktop control plane (`apps/desktop`).

- Sidebar grouped into Control / Account sections, wallet card anchored to the bottom of the rail
- Data-driven hero: connected agents, model count, settlement chain
- Model catalog: tinted capability chips (Reasoning / Vision / Agentic / Tools), larger type throughout
- Usage: live daily chart from the router's `/stats` with axis, gridlines, and per-day tooltips
- Settings: theme picker, copyable local endpoints
- Wallet: official Base and Solana marks, copyable addresses, an "Active" note replaces the disabled "Current default" button
- Funding dialog: Buy / Deposit tabs; Deposit shows both wallet addresses with copy buttons

## Beyond aesthetics

- The Deposit tab exposes both wallet addresses, which is new capability
- Copy buttons only show "Copied" once the clipboard write resolves; failures surface as a page notice or an inline error in the dialog
- The usage chart pads the router's traffic-only days into a continuous UTC window (at least 7 days, capped at 14) so quiet days render as zero bars and the "last N days" label is truthful
- Demo stats in `api.ts` now match the router's `/stats` shape

## Verification

- `npm run typecheck`, `vitest` (46 passed), `vite build`, and `prettier --check` are all clean
- Browser pass in both themes: axis labels sit within 1px of their gridlines, tooltips clamp inside the plot, and all three clipboard failure paths were exercised in a sandbox that denies writes
- Light-theme chip text measures 4.7 to 4.8:1 against its tint

## Notes for reviewers

- The sidebar sun/moon toggle is kept alongside the new Appearance picker
- Worth clicking one copy button in the packaged Electron app; the browser sandbox can only prove the failure path

## Summary by CodeRabbit

* **New Features**
  * Added interactive usage charts with daily request and cost data, hover details, and continuous date ranges.
  * Added theme selection, copyable local endpoints, and router connection status in Settings.
  * Added Buy and Deposit tabs with Solana deposit addresses.
  * Added wallet address copying, chain branding, active-network messaging, and refresh progress feedback.
  * Added clipboard confirmation states, updated dashboard metrics, and account-credit payment messaging.
  * Updated routing visuals with live agent and model counts.

* **Style**
  * Refined desktop layout, typography, spacing, animations, charts, dialogs, wallet cards, and model capability indicators.

### DIFF
--- a/apps/desktop/src/App.tsx
+++ b/apps/desktop/src/App.tsx
@@ -322,6 +322,10 @@ function Overview({
   const total = agents.length || 5;
   const modelCount = dashboard?.models.length ?? 0;
   const chain = dashboard?.proxy.configuredChain === "solana" ? "Solana" : "Base";
+  // A proxy authenticating with a BlockRun API key settles nothing on-chain, so
+  // name the account instead of claiming a chain that is not there.
+  const settlement =
+    dashboard?.proxy.authMode === "api-key" ? "paid with account credit" : `settling on ${chain}`;
   return (
     <>
       <section className="hero" aria-label="Routing status">
@@ -340,7 +344,7 @@ function Overview({
             </h2>
             <p>
               {online
-                ? `${modelCount ? compact(modelCount) : "—"} models available · settling on ${chain}`
+                ? `${modelCount ? compact(modelCount) : "—"} models available · ${settlement}`
                 : "Models, payment, and local agent connections stay in one place."}
             </p>
           </div>
@@ -1252,6 +1256,8 @@ function FundingDialog({
         <div className="funding-tabs" role="tablist" aria-label="Funding method">
           <button
             role="tab"
+            id="funding-tab-buy"
+            aria-controls="funding-pane"
             aria-selected={tab === "buy"}
             className={tab === "buy" ? "active" : ""}
             onClick={() => setTab("buy")}
@@ -1260,6 +1266,8 @@ function FundingDialog({
           </button>
           <button
             role="tab"
+            id="funding-tab-deposit"
+            aria-controls="funding-pane"
             aria-selected={tab === "deposit"}
             className={tab === "deposit" ? "active" : ""}
             onClick={() => setTab("deposit")}
@@ -1268,7 +1276,13 @@ function FundingDialog({
           </button>
         </div>
         {tab === "deposit" ? (
-          <div className="funding-pane" key="deposit">
+          <div
+            className="funding-pane"
+            key="deposit"
+            id="funding-pane"
+            role="tabpanel"
+            aria-labelledby="funding-tab-deposit"
+          >
             <div className="deposit-list">
               {depositRows.map((row) => (
                 <div className="deposit-row" key={row.chain}>
@@ -1322,7 +1336,13 @@ function FundingDialog({
             </p>
           </div>
         ) : (
-          <div className="funding-pane" key="buy">
+          <div
+            className="funding-pane"
+            key="buy"
+            id="funding-pane"
+            role="tabpanel"
+            aria-labelledby="funding-tab-buy"
+          >
             <div className="funding-amount">
               <label htmlFor="funding-usd">You pay</label>
               <div className="amount-input">
--- a/apps/desktop/src/App.tsx
+++ b/apps/desktop/src/App.tsx
@@ -1,4 +1,5 @@
 import { useEffect, useId, useMemo, useRef, useState } from "react";
+import { niceCeiling, normalizeStats, windowLabel, type UsageDay } from "./usage-stats.js";
 
 import type {
   AgentId,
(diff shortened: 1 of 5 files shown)


VickyXAI:
Reviewed. The refresh looks good and the copy-failure handling is more careful than most. But it adds three places where the UI states a number the router never reported, and none of them could be caught because the window math lives inside `App.tsx` — 1,451 lines and no test file touched. Fixed on `fix/367-control-plane` (branched off this PR's head); merge or cherry-pick `9078014`.

**1. The hero counts alias rows as models.** `const modelCount = dashboard?.models.length ?? 0` — `/v1/models` deliberately emits a row per shorthand, which is why the Models page runs the same array through `collapseModelAliases`. So the hero and `RoutingMap` advertise the raw length while the Models page right next to it reads "110 of 110", and the product's own copy says 76. Both now count the collapsed catalog.

**2. A window with no recent traffic draws a week that never happened.**

```ts
const firstKept = dates.map(parseDay).find((day) => day >= floor) ?? end;
const start = earliest(firstKept, shiftDays(end, -6));
```

When every day the router reported is older than `MAX_CHART_DAYS`, `find` returns undefined, `?? end` snaps to today, and the loop pads seven zero bars labelled "last 7 days" — next to a Requests card showing the full total. `getStats` reads the 7 most recent log *files*, which for an intermittent user span weeks, so this is not a corner case. There is no window to draw there, so it now draws none and the chart shows its empty state.

**3. The Requests card and the bars count different days.** The card reads `totalRequests` under a label describing the chart window. Those genuinely differ, for the same reason. It now sums the days it renders, so the number, the label and the bars cannot disagree.

Smaller, same theme:

- `peak` doubled as the divide-by-zero floor and as the number in the chart header, printing "peak 1 / day" over a window nobody routed in. The floor moved to the divisor.
- "Tokens routed" reads `inputTokens`. `AggregatedStats` has no token field and `/stats` has never sent one, so the card is permanently 0 in the packaged app while the demo shows 1.4M — the demo was hiding a dead metric. Replaced with savings against baseline, which the router does return, and the fake `inputTokens` is out of the demo.
- `healthLabel` returns "Connected" on `health === "ready"`, which per `adapters/shared.ts` only means installed + configured + proxy reachable. An agent still routing through its old config now reads "Restart pending". (CodeRabbit's App.tsx:1650 comment — it's right.)

The window math is now `src/usage-stats.ts` with 10 tests; two of them fail on this PR's logic.

**Two CodeRabbit comments are already fixed** by `bffe4a7` and can be resolved: the demo totals do derive from `demoDailyBreakdown` (474/6.35 reconcile across all three sources), and neither the `0.8fr 1.2fr` hero override nor `currentColor` is still in `styles.css`. **App.tsx:324 is a false positive** — the `authMode === "api-key"` branch is right there on the next line.

Verified: `tsc --noEmit` clean, 56 tests pass (was 46, none from this PR), `vite build` clean, merges cleanly with #371 in either order.

VickyXAI:
Correction to my review above: the `healthLabel` change was wrong and I've reverted it on the branch (`b86f8bf`).

CodeRabbit's App.tsx:1650 suggestion, which I followed, treats `restartRequired` as "a change is waiting". It isn't. `codex.ts` and `openclaw.ts` both set `restartRequired: configured`, so it is permanently true from the moment the agent connects — it means "this kind of agent needs a restart when you change it". Branching the label on it pins Codex and OpenClaw to "Restart pending" **forever**, which is worse than the "Connected" it replaced: a sometimes-stale label traded for an always-wrong one. `activationLabel` renders right beside it and already says "Restart gateway after changes" / "Restart app after changes", which is honest and doesn't decay.

So that CodeRabbit comment should be dismissed too, alongside App.tsx:324.

The real gap it was pointing at is still there and isn't fixable in this PR: nothing observes whether an agent process has actually picked the config up, so `restartRequired` can only ever describe a category, not a state. Worth an issue if you want the pill to mean something — @ramioca already flagged the same static-flag problem in #371's notes.

Everything else in the branch stands: `tsc` clean, 56 tests, `vite build` clean.

VickyXAI:
Pushed onto this branch as `ead500a` (maintainer edits were on) — the temporary `fix/367-control-plane` on origin is deleted, this PR is the single source now.

It is the two commits from my review squashed into one, with the `healthLabel` change **left out** for the reason in my correction above; a comment in the code now records why, so the next person doesn't re-apply CodeRabbit's suggestion. 56 tests pass (was 46), `tsc` and `vite build` clean.

VickyXAI (approved):
Re-reviewed on 2026-09-15 against current main (after #371/#379 landed): test-merged locally — tsc, 67 desktop tests, vite build and prettier all clean. Copy paths only ever carry public addresses and local endpoint URLs. Pushed `7e94516` to rewrite the `healthLabel` comment, which still argued against `restartRequired` after #379 deleted it.

PR #371, opened 2026-09-09
desktop: apply chain switches without a restart; Hermes needs none

ramioca:
## Summary

Two restart prompts in ClawRouter Desktop that users could not actually act on.

**Chain switch (Base ↔ Solana).** Switching wrote `~/.blockrun/.chain` and then said "Restart the ClawRouter/OpenClaw gateway", but the process that needed restarting is the proxy child Desktop itself spawned. The router resolves the chain once at startup and uses a different gateway per chain, so there is no in-process switch to call. Desktop now:

- restarts the proxy it owns (`ServiceSupervisor.restartProxy`: SIGTERM, wait for exit with a 5 s SIGKILL fallback, then the existing `ensureProxy` ownership check),
- confirms through `/health?full=true` that the new proxy signs on the requested chain,
- reports "Solana is active" only when both hold, and keeps the old message when the proxy on 8402 is not Desktop's child (an OpenClaw gateway or a terminal instance, which Desktop must not kill).

This matches what the Hermes plugin already does for its own `/clawrouter wallet` command.

**Hermes.** The card demanded a restart of the Hermes process after every change, forever, because the adapter's `restartRequired` mirrored `configured`. Hermes re-reads `config.yaml` whenever its mtime changes, resolves the provider key from `~/.hermes/.env` at call time, and Hermes Desktop starts a fresh `hermes --tui` process per chat. Nothing needs the process restarted; only an already-open session keeps its model until `/model`. The adapter now reports the change as live with that guidance.

## Verification

- `tsc --noEmit`, `vite build`, and `prettier --check` clean; vitest 53 passed (7 new: the four chain-switch outcomes, the CLI-failure path, the supervisor refusing a foreign proxy, and a real SIGTERM-and-relaunch of a spawned child).
- Live, driven over the DevTools protocol against the dev build with the packaged app quit:

| Switch | Time | Proxy pid | `/health` | `.chain` |
|---|---|---|---|---|
| Base → Solana | 2.0 s | 41043 → 41220 | solana | solana |
| Solana → Base | 1.0 s | 41220 → 41473 | base | base |

  Wallet badge stayed "Live", no restart hint appeared, and the proxy log shows a clean SIGTERM → "Proxy closed" → start on the new chain each time. The same round trip was repeated on top of `desktop-v0.1.3-preview.1` (pids 43056 → 43313 → 43388) where the hero and Hermes label update accordingly.

## Notes for reviewers

- The commit cherry-picks cleanly onto `desktop-v0.1.3-preview.1`; the electron and test files are identical between main and that tag.
- A switch clicked during the first-run runtime install, before the proxy exists, still gets the old message; the proxy then starts on the new chain anyway. Left alone as a first-launch-only case.
- Codex and OpenClaw still carry the static `restartRequired` flag, so on the 0.1.3 line they read "Configured · restart required" permanently after connecting. Out of scope here; happy to follow up.

## Summary by CodeRabbit

* **New Features**
  * Payment-chain changes now automatically restart the Desktop-managed proxy and verify the selected chain.
  * Hermes changes apply immediately to new chats, while open chats can switch models without restarting.
* **Bug Fixes**
  * Added clearer guidance when the proxy is externally managed, unhealthy, or reports a different chain.
  * Prevented unsuccessful chain commands from triggering unnecessary restarts.
  * Improved proxy restarts to stop lingering processes before relaunching and avoid restart hangs.

### DIFF
--- a/apps/desktop/electron/adapters/hermes.ts
+++ b/apps/desktop/electron/adapters/hermes.ts
@@ -10,7 +10,11 @@ export class HermesAdapter implements AgentAdapter {
   readonly id = "hermes" as const;
   readonly name = "Hermes";
   readonly description = "Hermes model-provider plugin backed by the local ClawRouter proxy.";
-  readonly activation = "restart-agent" as const;
+  // Hermes re-reads config.yaml whenever it changes and looks the provider key
+  // up in ~/.hermes/.env at call time, and Hermes Desktop starts a fresh
+  // `hermes --tui` process per chat. Nothing here needs the Hermes process
+  // restarted; only an already-open session keeps its model until /model.
+  readonly activation = "immediate" as const;
 
   managedPaths(context: AdapterContext): string[] {
     return [
@@ -30,10 +34,10 @@ export class HermesAdapter implements AgentAdapter {
       installed: Boolean(hermes),
       configured: hasHermesConfig(config),
       activation: this.activation,
-      restartRequired: hasHermesConfig(config),
+      restartRequired: false,
       details: hermes
         ? hasHermesConfig(config)
-          ? ["Restart the current Hermes process after changing this connection."]
+          ? ["New Hermes chats use ClawRouter; an open session switches with /model."]
           : []
         : ["Install Hermes before enabling the provider."],
     });
--- a/apps/desktop/electron/core/manager.ts
+++ b/apps/desktop/electron/core/manager.ts
@@ -169,11 +169,48 @@ export class ClawRouterManager {
           message: (result.stderr || result.stdout).trim() || `Could not switch to ${chain}.`,
         };
       }
+      const label = chain === "solana" ? "Solana" : "Base";
+      // The proxy reads the chain once at startup, so apply the switch by
+      // restarting the proxy Desktop launched instead of asking the user to
+      // restart something they never started. A proxy Desktop does not own is
+      // left alone and still needs its own restart.
+      let restarted = false;
+      try {
+        restarted = await this.supervisor.restartProxy();
+      } catch (error) {
+        return {
+          ok: false,
+          chain,
+          restartRequired: true,
+          message: `${label} selected, but the local proxy did not come back: ${
+            error instanceof Error ? error.message : String(error)
+          }. Restart ClawRouter Desktop to apply it.`,
+        };
+      }
+      if (!restarted) {
+        return {
+          ok: true,
+          chain,
+          restartRequired: true,
+          message: `${label} selected. Restart the ClawRouter/OpenClaw gateway to apply it.`,
+        };
+      }
+      const active = await this.activePaymentChain();
+      if (active !== chain) {
+        return {
+          ok: false,
+          chain,
+          restartRequired: true,
+          message: `${label} selected, but the restarted proxy reports ${
+            active ?? "no payment chain"
+          }. Check the proxy log in ClawRouter Desktop.`,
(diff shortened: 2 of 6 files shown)


VickyXAI:
Reviewed. The design is right — Desktop restarting the proxy it owns, and refusing to kill one it doesn't, is the correct shape, and the wrapper/grandchild test is the good kind of test. Three defects in the new restart path, all reachable from one chain switch. Fixed on `fix/371-supervisor` (branched off this PR's head); merge or cherry-pick `6cf70bc`.

**1. A signal-killed child reads as still running, and the restart then hangs forever.** `stopChild()` gates on `child.exitCode !== null`. Node leaves `exitCode` null for a process killed by a signal and reports the signal in `signalCode` — which is exactly how this file stops things:

```
after external SIGTERM -> exitCode: null  signalCode: SIGTERM  killed: false
guard `exitCode !== null` would RETURN EARLY? false
```

So the guard falls through on a dead child, `child.kill("SIGTERM")` no-ops against a closed handle, and `await exited` waits for an `"exit"` that already fired. `restartProxy()` never settles, and the chain switch spins with no way back. The added test reproduces it as a 5 s timeout. The same predicate is wrong in `liveChild()`, `childOwnsPort()` and `start()`; there is now one `hasExited()` they all use, and the wait is bounded so a child that outlives SIGKILL hands the decision to `stopProxy`'s port check instead of stalling.

**2. SIGKILL goes to a pid nobody re-confirmed.** `stopProxy()` captures `descendants` before the parent dies, then escalates to SIGKILL up to `STOP_GRACE_MS` later. A descendant that obeyed SIGTERM can have had its pid recycled inside that window, and the escalation only happens when *something* still holds 8402 — which by then may not be the pid in the list. It now re-reads the port (`listenersOnPort`, factored out of `listenersOwnedBy`) and kills only what is still holding it.

**3. The tracked child survives a failed stop.** `children.delete("proxy")` is skipped when `stopProxy()` throws, so a child Desktop already signalled stays in the map and the next `restartProxy()` tells the user Desktop does not own a proxy it started. Now in a `finally`. (This is CodeRabbit's supervisor.ts:117 comment — it's right.)

All three fail on this PR's code; `stopChild`'s failure is the hang itself.

**CodeRabbit's supervisor.ts:279 comment is a false positive.** It asks you to "terminate the owned process tree, or its listener descendants, and wait for port 8402 to close before calling `ensureProxy()`. Add a wrapper-child regression test." `stopProxy()` already does all of that, and `tests/supervisor.test.ts` already has the wrapper-child test. It reviewed `stopChild` without its caller.

Verified on this PR's head plus the fix: `tsc --noEmit` clean, 57 tests pass, and it merges cleanly with #367 in either order.

VickyXAI:
Pushed onto this branch as `998874e` (maintainer edits were on) — the temporary `fix/371-supervisor` on origin is deleted, this PR is the single source now. 57 tests pass, `tsc` clean.

VickyXAI (approved):
The supervisor work holds up: `hasExited` fixes a real bug (`exitCode` stays null for a signal-killed child, which is exactly how this file stops things), the descendant SIGKILL is re-confirmed against the live port so a recycled pid can't be hit, and a proxy on 8402 that isn't Desktop's child is left alone. Live chain-switch table + 7 new tests are enough. Merging ahead of #379, which deletes the `restartRequired` field this PR sets — same lines, so this one goes first.

